PR #14, opened 2016-01-27
Addressing an incompatibility between browsers when reporting errors.

MSLaguana:
Also addressing a couple of tslint issues.

The main problem here was that chrome reports errors in one way, with an `exceptionDetails.text` property, while iOS and android have a `result.description` property instead.

### DIFF
--- a/debugger/webkit/webKitDebugAdapter.ts
+++ b/debugger/webkit/webKitDebugAdapter.ts
@@ -465,11 +465,11 @@ export class WebKitDebugAdapter implements IDebugAdapter {
                     // In these cases, return a dummy stack frame
                     return {
                         id: i,
-                        name: "Unknown",
+                        name: 'Unknown',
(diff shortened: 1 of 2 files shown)


debugger/webkit/webKitDebugAdapter.ts
```diff
@@ -465,11 +465,11 @@ export class WebKitDebugAdapter implements IDebugAdapter {
                     // In these cases, return a dummy stack frame
                     return {
                         id: i,
-                        name: "Unknown",
+                        name: 'Unknown',
```
nisheetjain: why do we need this change?

MSLaguana: These minor changes are fixing tslint issues that I hadn't noticed earlier. The important fix is entirely in the `evaluate` method below

nisheetjain:
:+1:

PR #16, opened 2016-01-27
Fixing an issue where we would not attach to iOS applications with ch…

MSLaguana:
…aracters in its name which require URL encoding.


meenakb:
:+1:

nisheetjain:
:+1:

Issue #26, opened 2016-01-29
Debug adapter process has terminated unexpectedly


aitorllj93:
On Linux Mint 17.2

Also, the Cordova:Run and Cordova:Build commands return
```
No handler found for the command: 'cordova.run'. Ensure there is an activation event defined, if you are an extension.
```

The cordova application was succesful created an built from command line for android platform. Only fails on vscode.

## launch.json
```json
{
	"version": "0.2.0",
	"configurations": [
         {
            "name": "Cordova",
            "type": "cordova",
            "request": "launch",
            "platform": "android",
            "target": "device"
        }
	]
}
```

## config.xml
```xml
<?xml version='1.0' encoding='utf-8'?>
<widget id="com.test.package" version="1.0.0" xmlns="http://www.w3.org/ns/widgets" xmlns:cdv="http://cordova.apache.org/ns/1.0">
    <name>gxdcalidad</name>
    <description>
        A sample Apache Cordova application that responds to the deviceready event.
    </description>
    <author email="dev@cordova.apache.org" href="http://cordova.io">
        Apache Cordova Team
    </author>
    <content src="index.html" />
    <access origin="*" />
    <preference name="Fullscreen" value="true" />
    <preference name="DisallowOverscroll" value="true" />
    <engine name="android" spec="~4.1.1" />
    <plugin name="com.lampa.startapp" spec="~0.0.5" />
    <plugin name="cordova-plugin-whitelist" spec="1" />
    <allow-navigation href="*" />
    <allow-intent href="http://*/*" />
    <allow-intent href="https://*/*" />
    <allow-intent href="tel:*" />
    <allow-intent href="sms:*" />
    <allow-intent href="mailto:*" />
    <allow-intent href="geo:*" />
    <platform name="android">
        <allow-intent href="market:*" />
    </platform>
    <platform name="ios">
        <allow-intent href="itms:*" />
        <allow-intent href="itms-apps:*" />
    </platform>

</widget>
```

## Distro details

```
::::::::::::::
/etc/lsb-release
::::::::::::::
DISTRIB_ID=LinuxMint
DISTRIB_RELEASE=17.2
DISTRIB_CODENAME=rafaela
DISTRIB_DESCRIPTION="Linux Mint 17.2 Rafaela"
::::::::::::::
/etc/os-release
::::::::::::::
NAME="Ubuntu"
VERSION="14.04.3 LTS, Trusty Tahr"
ID=ubuntu
ID_LIKE=debian
[...]

PR #27, opened 2016-01-29
Fixing incorrect case in file paths

MSLaguana:
While I'm here, also updating some imports to the new syntax.

I believe that this should fix issue #26 as these were the only two paths with incorrect casing that I could see.


meenakb:
Those are the two occurences I see as well. :+1:

PR #31, opened 2016-02-02
Fixing over-encoding of paths for iOS device debugging

### LINKED ISSUE #30: [cordova-debug-adapter] - Error: Unable to find iOS target device/simulator
Hey there, 

iOS 8.1.2, el capitan. 

Tried out the tools after a clean install of cordova tools. The simulator works flawless, but I am running into some issues on debugging on a device. (iOS)

The "Run iOS on device" threw an error on "Configuring debugging proxy" step:
"[cordova-debug-adapter] Error: Unable to find iOS target device/simulator. Try specifying a different "port" parameter in launch.json"

The port is set to **9220** in the **launch.json**. I was able to get rid of the error message by restarting visual studio code and changing the port to 9221. However, I am still not able to hit a breakpoint or type anything into the debug console. (i.e. console.log('test'); -> "TypeError: Cannot read property 'runting_evaluate' of undefined). 
Visual Studio Code seems to be stuck of some sort. Or at least the loading bar in the top left corner indicates a busy state. 

https://github.com/google/ios-webkit-debug-proxy; I find only references to 9221 - 9322. 

Am I missing something, or is this a bug?

**Update:** Able to reproduce on device using iOS 9.2.1 as well.

MSLaguana:
When attempting to determine which webview to attach to we URL encode the package name to deal with spaces and other special characters. However prior to this fix we pass the whole on-device path for device debugging, and we would URL encode the path separators. Now we only match on, and encode, the app folder itself.

Fixes #30

### DIFF
--- a/src/debugger/cordovaDebugAdapter.ts
+++ b/src/debugger/cordovaDebugAdapter.ts
@@ -231,7 +231,9 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
         this.outputLogger('Configuring debugging proxy');
         return CordovaIosDeviceLauncher.startWebkitDebugProxy(attachArgs.port, attachArgs.webkitRangeMin, attachArgs.webkitRangeMax).then(() => {
             if (attachArgs.target.toLowerCase() === 'device') {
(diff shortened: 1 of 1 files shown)


nisheetjain:
:+1:

joshuaskelton:
Looks good to me.

PR #34, opened 2016-02-03
Telemetry opt in

meenakb:


README.md
```diff
@@ -85,3 +85,31 @@ Intellisense helps you discover objects, functions, and parameters in libraries
...
+
+### OS X / Linux
+
+* Close VS Code
+* Open the terminal
+* Type cd ~/.vscode-cordova
+* Type vi VSCodeTelemetrySettings.json
+* Replace "optIn":true with "optIn":false
+
+Save the file via Esc ZZ. Collection of usage data should now be disabled.
```
nisheetjain: I think you mean ":wq" to save & exit in Vi. IMO you can skip the Vi/notepad details and keep instructions short. Devs know how to open a file in their "editor"

src/utils/telemetry.ts
```diff
@@ -213,7 +213,7 @@ export module Telemetry {
                     TelemetryUtils.sessionId = TelemetryUtils.generateGuid();
                     TelemetryUtils.userType = TelemetryUtils.getUserType();
 
-                    Telemetry.isOptedIn = isOptedInValue;
+                    Telemetry.isOptedIn = TelemetryUtils.getTelemetryOptInSetting();
```
nisheetjain: I am afraid that we may need more work here or change instructions. if we are not dropping a telemetrySettings.json file, user will not find a file with optIn:true (which is fine). We can ask user to add entry for `optIn: false`
if we do that we can get rid of  isOptedIn from the whole call chain because it's mostly settings.json which is driving the telemetry opt in status and not Telemetry.init

meenakb: Changed the language per your suggestion.

README.md
```diff
@@ -85,3 +85,21 @@ Intellisense helps you discover objects, functions, and parameters in libraries
...
+
+If you don’t wish to send usage data to Microsoft, please follow the instructions below to disable its collection.
+
+Important Notice: You will need to apply these changes after every update to disable collection of usage data. These changes do not survive product updates.
+
+### Windows
+
+* Edit VSCodeTelemetrySettings.json file at %appdata%\vscode-cordova and add "optIn":false.
+
+Save the file via CTRL+S and exit Notepad. Restart VS Code. Collection of usage data should now be disabled.
```
nisheetjain: this line can be removed...

nisheetjain:
:+1:

PR #44, opened 2016-02-22
Adding 'private' to package.json to avoid accidental npm publish

MSLaguana:


digeff:
:+1: as long as we know for sure this won't mess anything in the marketplace

MSLaguana:
It hasn't caused any issues for the react-native extension, so it should be fine. The only meaning that I'm aware of is to not allow you to `npm publish` a package.

digeff:
Great! Go for it!

meenakb:
:package: :+1:

Issue #48, opened 2016-03-06
[Issue] [Cordova Tool] Unable to debug with error: "Path must be string. Received null" after upgrade VS Code from 0.10.9 to 0.1010


saravmajestic:
I am getting the error `[cordova-debug-adapter] TypeError: Path must be string. Received null` when trying to debug my phonegap app using "Cordova tools".
Steps:

Installed VSCode Version 0.10.9 (0.10.9)
installed cordova extension from extensions gallery
selected Cordova from debug configuration and used default `launch.json`
clicked 'Run Android on emulator' Got the error as above.

I tried to add the workaround as mentioned in [https://github.com/Microsoft/vscode/issues/3625](https://github.com/Microsoft/vscode/issues/3625)
I tried to set sourceMaps as false and "outDir": "${workspaceRoot}" , but still getting same error.

I don't know which sourcemaps I have to specify as mentioned in the workaround. I am using Mac 10.10.5

Screenshot as below:
<img width="1440" alt="screen shot 2016-03-06 at 5 46 28 pm" src="https://cloud.githubusercontent.com/assets/1147025/13554046/88eb3df6-e3c3-11e5-8a56-55f1487ba433.png">

DickvdBrink:
Same issue here (with the insider build)

MSLaguana:
What version of node do you have installed? Looks like this might be to do with a change in node 5.7

saravmajestic:
My Node version is pretty old 0.10.32. Should I upgrade to latest?

MSLaguana:
That is quite old, but shouldn't be related to this particular issue then. I'll investigate further.

digeff:
@saravmajestic Could you please try:
1. Setting sourceMaps: false and deleting the outDir key
2. Setting sourceMaps: true and "outDir": "${workspaceRoot}"
3. Setting sourceMaps: true and "outDir": "${workspaceRoot}/www" or whatever is the path to your index.html

saravmajestic:
@digeff I tried all these combinations now and still getting the error.

My phonegap version is 5.3.9. I will try to generate a new cordova project and try this out. In the meantime, let me know if you guys have any other work around

MSLaguana:
For a workaround, you should be able to specify `"sourceRoot": "${workspaceRoot}/www"` I believe. If that does fix it, please let us know and we'll put in a proper fix.

MSLaguana:
[...]

PR #50, opened 2016-03-08
Update default launch.json parameters

MSLaguana:
vscode 0.10.10 changed the debug adapter communication protocol, and no longer adds the cwd parameter that we depend on.

### DIFF
--- a/package.json
+++ b/package.json
@@ -165,7 +165,7 @@
                             "cwd": {
                                 "type": "string",
                                 "description": "The root of the project"
-                            }
+                            },
                             "target": {
                                 "type": "string",
                                 "description": "either 'device', 'emulator', or identifier for a specific device / emulator",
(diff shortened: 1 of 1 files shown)


MSLaguana:
This fixes #48 for new projects. Projects which already have a launch.json will need to manually make the fix.

package.json
```diff
@@ -140,19 +147,25 @@
...
                         ],
                         "properties": {
                             "platform": {
                                 "type": "string",
                                 "description": "The platform to run on"
                             },
+                            "cwd": {
+                                "type": "string",
+                                "description": "The root of the project"
+                            }
```
DickvdBrink: Missing comma, so this is invalid JSON.

joshuaskelton: Missing a comma here.

joshuaskelton:
Looks good to me! :+1:

DickvdBrink:
What if you added `vscode.workspace.rootPath`
[here](https://github.com/Microsoft/vscode-cordova/blob/4439865650b94f8701e5645294af71aa29855ed2/src/debugger/cordovaDebugAdapter.ts#L35) and [here](https://github.com/Microsoft/vscode-cordova/blob/4439865650b94f8701e5645294af71aa29855ed2/src/debugger/cordovaDebugAdapter.ts#L64)
That way it would be fixed for users without changing the `launch.json` because they can't really know that

MSLaguana:
Unfortunately we can't do that: the way that debuggers are implemented is you create a "debug adapter" process which is a separate executable that communicates with vscode proper via a text protocol. This debug adapter process has no access to vscode APIs directly, the only information it has is passed via the launchArgs (or other arguments for other messages).

PR #52, opened 2016-03-08
Merging vscode-chrome-debug/master to keep the histories similar

MSLaguana:


digeff:
I think that HTTP goes in uppercase, but apart from that this PR is awesome! :+1: :+1: :+1:

PR #74, opened 2016-03-25
Update Telemetry to Use VS Code Extension Telemetry Package

joshuaskelton:
## Summary

Removed the Application Insights package and replaced it with VS Code Extension Telemetry. This change required us to have the debug process send it's telemetry back to the extension process.

## Changes

- Removed AppInsights dependency
- Added vscode-extension-telemetry dependency
- Ensured our telemetry data points are the format that vscode-extension-telemetry expects
- Ported extension messaging from React Native tools


package.json
```diff
@@ -275,12 +275,12 @@
         "test": "node ./node_modules/mocha/bin/mocha --recursive -u tdd ./out/debugger/test/ ./out/test/debugger"
     },
     "dependencies": {
-        "applicationinsights": "^0.15.8",
         "elementtree": "^0.1.6",
-        "getmac": "1.0.7",
         "plist-with-patches": "^0.5.1",
         "q": "^1.4.1",
+        "rimraf": "^2.5.0",
```
digeff: Remember to add it to ThirdPartyNotices.txt and follow the required procedure.

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as Q from "q";
+import * as net from "net";
+
+export let ErrorMarker = "vscode-cordova-error-marker";
+
+/**
+ * Defines the messages sent to the extension.
+ * Add new messages to this enum.
+ */
+export enum ExtensionMessage {
```
digeff: I'd move the ExtensionMessage to another file, and all the cordova specific things, so then we can use the exact same extensionMessaging.ts file in both vscode-react-native and vscode-cordova.

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,132 @@
...
+import {
+    ErrorMarker,
+    ExtensionMessage,
+    ExtensionMessageSender,
+    MessageWithArguments
+} from "../common/extensionMessaging";
+
+import {Telemetry} from "../utils/telemetry";
+
+export class ExtensionServer implements vscode.Disposable {
```
digeff: Same as ExtensionMessaging. Can we move all the individual messages and cordova specific logic to another file, so we can share this exact same file with vscode-react-native?

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Sends messages to the extension.
+ */
+export class ExtensionMessageSender {
+    public static getExtensionPipePath(): string {
+        switch (process.platform) {
+            case "win32":
+                return "\\\\?\\pipe\\vscodecordova";
+            default:
+                return "/tmp/vscodecordova.sock";
```
digeff: Should we use os.tmpdir() here ? the actual $TMPDIR in OS X is a subfolder of /tmp
Do all the users have permission to write to /tmp? Do they need it for a pipe there?

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,132 @@
...
+    private serverInstance: net.Server = null;
+    private messageHandlerDictionary: { [id: number]: ((...argArray: any[]) => Q.Promise<any>) } = {};
+    private pipePath: string;
+    private telemetryReporterDictionary: {} = {};
```
digeff: Where is this used?

src/utils/telemetry.ts
```diff
@@ -387,4 +349,47 @@ export module Telemetry {
...
+                this.extensionMessageSender = new ExtensionMessageSender();
+            }
+
+            sendTelemetryEvent(eventName: string, properties?: ITelemetryEventProperties, measures?: ITelemetryEventMeasures) {
+                this.extensionMessageSender.sendMessage(ExtensionMessage.SEND_TELEMETRY, [this.extensionId, this.extensionVersion, this.appInsightsKey, eventName, properties, measures])
+                .catch(function(){});
+            }
+        }
+
+        export function sendExtensionTelemetry(extensionId: string, extensionVersion: string, appInsightsKey: string, eventName: string, properties: ITelemetryEventProperties, measures: ITelemetryEventMeasures): void {
```
digeff: This function doesn't seem to be cohesive with this module. I'd move it to another file/module/class

src/utils/telemetry.ts
```diff
@@ -104,8 +107,27 @@ export module Telemetry {
                         (<TelemetryActivity> event).end();
                     }
 
-                    if (appInsights.client) { // no-op if telemetry is not initialized
-                        appInsights.client.trackEvent(event.name, event.properties);
+                    if (Telemetry.reporter) {
```
digeff: I'd print at least a warning if they call .send but we don't have a telemetry reporter configured

src/utils/telemetry.ts
```diff
@@ -104,8 +107,27 @@ export module Telemetry {
...
                     }
 
-                    if (appInsights.client) { // no-op if telemetry is not initialized
-                        appInsights.client.trackEvent(event.name, event.properties);
+                    if (Telemetry.reporter) {
+                        var properties: ITelemetryEventProperties = {};
+                        var measures: ITelemetryEventMeasures = {};
+
+                        for (var key in event.properties) {
+                            switch (typeof event.properties[key]) {
```
digeff: You use this 4 times. I'd instead add:
var value = event.properties[key];

src/utils/telemetry.ts
```diff
@@ -173,39 +187,20 @@ export module Telemetry {
...
-                    .start().client;
-                appInsights.client.config.maxBatchIntervalMs = 100;
-                sender.WAIT_BETWEEN_RESEND = 0;
-                telemetryLogger.disableWarnings = true;
-
-                if (client && client.context && client.context.keys && client.context.tags) {
-                    // Remove potential PII
-                    let machineNameKey = client.context.keys.deviceMachineName;
-                    client.context.tags[machineNameKey] = '';
+                if (initOptions.isExtensionProcess) {
```
digeff: Instead of passing initOptions.isExtensionProcess as a parameter, I'd pass the Telemetry.reporter. I think it'll be a better design.

digeff: Instead of passing initOptions.isExtensionProcess as a parameter, I'd pass the Telemetry.reporter. I think it'll be a better design.

package.json
guillaumejenkins: I doubt we'll have OSS approval in time for tomorrow... We might want to discuss this

package.json
```diff
@@ -275,12 +275,12 @@
...
         "source-map": "^0.5.3",
+        "vscode-extension-telemetry": "0.0.5",
```
guillaumejenkins: This might need to go in the OSS request as well

src/utils/telemetry.ts
```diff
@@ -86,10 +85,14 @@ export module Telemetry {
...
 
-        export function init(appNameValue: string, appVersion?: string, isOptedInValue?: boolean): Q.Promise<any> {
+        export interface ITelemetryInitOptions {
+            isExtensionProcess: boolean;
+        }
+
+        export function init(appNameValue: string, appVersion: string, initOptions: ITelemetryInitOptions): Q.Promise<any> {
             try {
                 Telemetry.appName = appNameValue;
-                return TelemetryUtils.init(appVersion, isOptedInValue);
```
digeff: You should remove methods setTelemetryOptInSetting getTelemetryOptInSetting and all other unused methods/code

src/utils/telemetry.ts
```diff
@@ -387,4 +349,47 @@ export module Telemetry {
...
+
+        export interface ITelemetryEventMeasures {
+            [key: string]: number;
+        }
+
+        export interface ITelemetryReporter {
+            sendTelemetryEvent(eventName: string, properties?: ITelemetryEventProperties, measures?: ITelemetryEventMeasures);
+        }
+
+        class ExtensionTelemetryReporter implements ITelemetryReporter {
```
digeff: I'd change the name to:
TelemetrySentThorughExtensionReporter

src/utils/telemetry.ts
```diff
@@ -387,4 +349,47 @@ export module Telemetry {
...
+            constructor(extensionId: string, extensionVersion: string, key: string) {
+                this.extensionId = extensionId;
+                this.extensionVersion = extensionVersion;
+                this.appInsightsKey = key;
+                this.extensionMessageSender = new ExtensionMessageSender();
+            }
+
+            sendTelemetryEvent(eventName: string, properties?: ITelemetryEventProperties, measures?: ITelemetryEventMeasures) {
+                this.extensionMessageSender.sendMessage(ExtensionMessage.SEND_TELEMETRY, [this.extensionId, this.extensionVersion, this.appInsightsKey, eventName, properties, measures])
+                .catch(function(){});
```
digeff: You should call .done()

src/utils/telemetry.ts
```diff
@@ -387,4 +349,47 @@ export module Telemetry {
...
+        export interface ITelemetryEventProperties {
+            [key: string]: string;
+        }
+
+        export interface ITelemetryEventMeasures {
+            [key: string]: number;
+        }
+
+        export interface ITelemetryReporter {
+            sendTelemetryEvent(eventName: string, properties?: ITelemetryEventProperties, measures?: ITelemetryEventMeasures);
```
digeff: I think it'd be better to make these arguments required.
Extra points if you use a decorator xD!

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+
+    public sendMessage(message: ExtensionMessage, args?: any[]): Q.Promise<any> {
+        let deferred = Q.defer<any>();
+        let messageWithArguments: MessageWithArguments = { message: message, args: args };
+        let body = "";
+
+        let pipePath = ExtensionMessageSender.getExtensionPipePath();
+        let socket = net.connect(pipePath, function() {
```
guillaumejenkins: If we are connecting to the pipe each time we send a message, should we be closing it each time as well?

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Starts the server.
+     */
+    public setup(): Q.Promise<void> {
+
+        let deferred = Q.defer<void>();
+
+        let launchCallback = (error: any) => {
+            // Log.logInternalMessage(LogLevel.Info, "Extension messaging server started.");
```
meenakb: remove?

src/utils/telemetry.ts
MSLaguana: I disagree; this would significantly clutter up output in tests, where we don't bother to configure telemetry. I think that a lack of reporter should be equivalent to opting out of telemetry, and have send just be a no-op.

src/extension/extensionServer.ts
joshuaskelton: Will do!

src/utils/telemetry.ts
joshuaskelton: I'm just mimicking the interface exposed by vscode-extension-telemetry, which has both as optional parameters.

src/utils/telemetry.ts
joshuaskelton: Good catch.

src/utils/telemetry.ts
joshuaskelton: Hmm. Seems too wordy to me.

src/utils/telemetry.ts
joshuaskelton: I'll remove setTelemetryOptInSetting, as it is not used. I'll keep getTelemetryOptInSetting because it is used, and allows the user to opt out by changing the setting in a json file.

package.json
joshuaskelton: This is a MS owned package. Is this necessary?

src/extension/extensionServer.ts
joshuaskelton: The telemetryReporterDictionary is used in sendExtensionTelemetry. We build an associative array of reporters for each telemetry source (extension, debugger, etc).

src/common/extensionMessaging.ts
```diff
@@ -0,0 +1,69 @@
...
+            body += data;
+        });
+
+        socket.on("error", function(data: any) {
+            deferred.reject(new Error("An error ocurred while handling message: " + ExtensionMessage[message]));
+        });
+
+        socket.on("end", function() {
+            try {
+                if (body === ErrorMarker) {
```
guillaumejenkins: Are we guaranteed that errors will be reported as exactly the error marker string? What about newlines / white space / additional text? Should we instead be using indexOf() !== -1?

package.json
guillaumejenkins: If the telemetry package is MIT, then _maybe_ we don't need to do anything; if it isn't, we most certainly will need to do something. In any case, I think it's worth a quick email to the right people.

package.json
joshuaskelton: Done.

src/utils/telemetry.ts
joshuaskelton: All that would do is move this check to Telemetry.init. Is that really a better design?

src/utils/telemetry.ts
digeff: I think we should configure a "null" reporter for the tests, and maybe even gather telemetry and check that we are configuring the right telemetry.

package.json
joshuaskelton: It is MIT licensed. But I need to add it to the third party notices.

src/utils/telemetry.ts
digeff: I'd pass the reporter to telemetry.init too, so then we can pass a null reporter or inmemoryreporter for the tests...

src/utils/telemetry.ts
digeff: UsePipeReporter? RemotePipeReporter?

ExtensionTelemetryReporter  makes me think this is the reporter for the Extension, but this is actually the reporter for everything that is not the extension.

src/utils/telemetry.ts
digeff: Ah, then it makes sense to keep it as-is.

src/extension/extensionServer.ts
joshuaskelton: Ah, was conflating this with Telemetry.reporterDictionary. Will remove.

ThirdPartyNotices.txt
```diff
@@ -26,15 +26,15 @@ This project incorporates components from the projects listed below. The origina
 
 %% Microsoft/ApplicationInsights-node.js NOTICES AND INFORMATION BEGIN HERE
```
meenakb: If the changes to this file were not provided by Riya, please get this reviewed by her :)

meenakb: Also, keep in mind that we will need to add notices to transitive dependencies as well.

src/common/extensionMessaging.ts
joshuaskelton: @dlebu Might know better?

hamiltonia: as with the previous, these platform checks should be moved to the platform helper if possible

src/common/extensionMessaging.ts
dlebu: We are creating a temporary file, not a temporary directory. I think this location is fine, if you open the /tmp dir on your mac, you will see that chrome, vscode and others use that location for temporary files and folders.

src/common/extensionMessaging.ts
dlebu: @guillaumejenkins @joshuaskelton The server is in charge of handling the lifetime of the socket. It will close the channel after a response has been sent with the result of the requested operation.

src/common/extensionMessaging.ts
dlebu: [This line](https://github.com/Microsoft/vscode-cordova/pull/74/files#diff-a0bb3c3c1809e7b5cf209c8263767d8fR89) handles that. Yes, we should be getting the marker only.

ThirdPartyNotices.txt
```diff
@@ -26,15 +26,15 @@ This project incorporates components from the projects listed below. The origina
...
 =========================================
-The MIT License (MIT) 
+The MIT License (MIT)
 
-Copyright � 2014 Microsoft Corporation
+Copyright � 2014 Microsoft Corporation
```
guillaumejenkins: We need to fix the copyright symbols, the quotation marks and some of the hyphens

ThirdPartyNotices.txt
```diff
@@ -313,6 +313,26 @@ IN THE SOFTWARE.
...
+
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR
+IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+=========================================
+isaacs/rimraf Working Draft NOTICES AND INFORMATION
```
guillaumejenkins: This should be "END OF ..."

src/common/extensionMessaging.ts
guillaumejenkins: @dlebu Thanks for the clarification!

src/common/extensionMessaging.ts
guillaumejenkins: @dlebu Thanks for the clarification!

ThirdPartyNotices.txt
joshuaskelton: Removed rimraf. vscode-extension-telemetry depends on applicationinsights and winreg, so we are okay on that end.

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,130 @@
...
+    };
+
+    /**
+     * Recovers the server in case the named socket we use already exists, but no other instance of VSCode is active.
+     */
+    private recoverServer(error: any): void {
+        let errorHandler = (e: any) => {
+            /* The named socket is not used. */
+            if (e.code === "ECONNREFUSED") {
+                CordovaProjectHelper.deleteDirectoryRecursive(this.pipePath);
```
dlebu: Any reason for which this call changed to sync from async?

joshuaskelton: I didn't change it? It was sync to begin with. Should it be async?

dlebu: That's how it looks like over in the [react-native repo](https://github.com/Microsoft/vscode-react-native/blob/master/src/extension/extensionServer.ts#L168).

src/utils/telemetry.ts
```diff
@@ -173,39 +189,20 @@ export module Telemetry {
...
+                else {
+                    Telemetry.reporter = new ExtensionTelemetryReporter(Telemetry.appName, appVersion, TelemetryUtils.APPINSIGHTS_INSTRUMENTATIONKEY);
                 }
 
-                // Change endpoint to match Aimov key
-                client.config.endpointUrl = "https://vortex.data.microsoft.com/collect/v1";
-
-                return Q.all([TelemetryUtils.getUserId(), TelemetryUtils.getMachineId()])
-                .spread<any>(function (userId: string, machineId: string): void {
+                return Q.all([TelemetryUtils.getUserId()])
```
dlebu: Nit: No need for `Q.all` here anymore. You can use `TelemetryUtils.getUserId().then(...)` directly.

src/extension/extensionServer.ts
```diff
@@ -0,0 +1,130 @@
...
+            this.serverInstance.close();
+            this.serverInstance = null;
+        }
+    }
+
+    /**
+     * Sends telemetry
+     */
+    private sendTelemetry(extensionId: string, extensionVersion: string, appInsightsKey: string, eventName: string, properties: {[key: string]: string}, measures: {[key: string]: number}): Q.Promise<any> {
+        Telemetry.sendExtensionTelemetry(extensionId, extensionVersion, appInsightsKey, eventName, properties, measures);
```
dlebu: I would make `sendExtensionTelemetry` asynchronous and return its promise here instead of a new object.

src/utils/telemetry.ts
```diff
@@ -387,4 +340,48 @@ export module Telemetry {
...
+            private appInsightsKey: string;
+
+            constructor(extensionId: string, extensionVersion: string, key: string) {
+                this.extensionId = extensionId;
+                this.extensionVersion = extensionVersion;
+                this.appInsightsKey = key;
+                this.extensionMessageSender = new ExtensionMessageSender();
+            }
+
+            sendTelemetryEvent(eventName: string, properties?: ITelemetryEventProperties, measures?: ITelemetryEventMeasures) {
```
dlebu: Does it make sense to make this asynchronous as well and return its result directly in the `sendExtensionTelemetry` method?

src/extension/extensionServer.ts
joshuaskelton: Sorry, was looking at the wrong line. Will address.

src/utils/telemetry.ts
joshuaskelton: Will fix.

PR #86, opened 2016-03-31
For iOS device, create the .ipa from the .app if it wasn't created

guillaumejenkins:
cordova-ios 3.8.0 and older doesn't generate a .ipa on build. Our extension looks specifically for a .ipa, so we weren't compatible with these older cordova-ios releases. Now, we generate the .ipa from the .app if Cordova didn't create the .ipa.

@MSLaguana

### DIFF
--- a/src/debugger/cordovaDebugAdapter.ts
+++ b/src/debugger/cordovaDebugAdapter.ts
@@ -243,7 +243,7 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
                     let ipaFiles = files.filter((file) => /\.ipa$/.test(file));
 
                     if (ipaFiles.length !== 0) {
-                        return Q.resolve(path.join(buildFolder, ipaFiles[0]));
+                        return path.join(buildFolder, ipaFiles[0]);
                     }
 
                     // No .ipa was found, look for a .app to convert to .ipa using xcrun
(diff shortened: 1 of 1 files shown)


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -241,11 +241,27 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
...
+                    if (appFiles.length === 0) {
+                        throw new Error('Unable to find a .app or a .ipa to install');
                     }
-                    let ipaFile = path.join(buildFolder, ipaFiles[0]);
 
+                    let appFile = path.join(buildFolder, appFiles[0]);
+                    let ipaFile = path.join(buildFolder, path.basename(appFile, path.extname(appFile)) + '.ipa'); // Convert [path]/foo.app to [path]/foo.ipa
+
+                    return execCommand('xcrun', ['-v', '-sdk', 'iphoneos', 'PackageApplication', `${appFile}`, '-o', `${ipaFile}`], errorLogger).then(() => {
+                        return Q.resolve(ipaFile);
```
MSLaguana: No need for Q.resolve here, it's already in a promisy context

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -241,11 +241,27 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
...
+                    }).catch((err): string => {
+                        throw new Error('Error converting .app to .ipa');
```
MSLaguana: Perhaps include the name of the .app in the error?

MSLaguana:
Looks good to me.

src/debugger/cordovaDebugAdapter.ts
guillaumejenkins: Done

src/debugger/cordovaDebugAdapter.ts
guillaumejenkins: Done

PR #89, opened 2016-04-04
Small tooltip clarifications in package.json for launch configurations

guillaumejenkins:
Should help with issue #88

### DIFF
--- a/package.json
+++ b/package.json
@@ -195,7 +195,7 @@
                         "properties": {
                             "platform": {
                                 "type": "string",
-                                "description": "The target platform to run for (either 'ios' or 'android', Windows is not supported yet)"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -195,7 +195,7 @@
                         "properties": {
                             "platform": {
                                 "type": "string",
-                                "description": "The platform to run on"
+                                "description": "The target platform to run for (either 'ios' or 'android', Windows is not supported yet)"
                             },
```
digeff: There are a lot of platforms we don't support. I'd either mention all of them or none.
Also, yet implies kindofish that we are going to support it soon.

guillaumejenkins: You are right; I changed it a bit

digeff:
:+1:

PR #94, opened 2016-04-06
Bug Fix: Build Error Unreachable Code detected

paulpatarinski:
Removing the usage of both throw & break which caused a build error


digeff:
:+1:

hamiltonia:
Thanks!

Issue #95, opened 2016-04-06
Build Error : Unreachable code detected.


paulpatarinski:
Building the extension results in the following error : 

debugger\common\v8Protocol.ts(215,5): error TS7027: Unreachable code detected.

Looks like there the default condition has a throw & break 

```
default:
	throw new Error('Unknown state');
	break;
```

Removing the break fixes that issue (PR Pending #94), but results in a different error #93

Issue #93, opened 2016-04-06
Build Error : Type Not assignable


paulpatarinski:
Building the extension results in the following error : 

debugger\webkit\webKitDebugAdapter.ts(348,16): error TS2322: Type 'Promise<Promise<SetBreakpointByUrlResponse>[]>' is not assignable to type 'Promise<SetBreakpointByUrlResponse[]>'.

Stack Trace : 

[08:16:10] Starting 'build'...
debugger\webkit\webKitDebugAdapter.ts(348,16): error TS2322: Type 'Promise<Promise<SetBreakpointByUrlResponse>[]>' is not assignable to type 'Promise<SetBreakpointByUrlResponse[]>'.
  Type 'Promise<SetBreakpointByUrlResponse>[]' is not assignable to type 'SetBreakpointByUrlResponse[]'.
    Type 'Promise<SetBreakpointByUrlResponse>' is not assignable to type 'SetBreakpointByUrlResponse'.
      Property 'result' is missing in type 'Promise<SetBreakpointByUrlResponse>'.
debugger\webkit\webKitDebugAdapter.ts(512,35): error TS2339: Property 'error' does not exist on type 'Promise<GetPropertiesResponse>'.
debugger\webkit\webKitDebugAdapter.ts(513,34): error TS2339: Property 'result' does not exist on type 'Promise<GetPropertiesResponse>'.
src\utils\telemetryHelper.ts(86,13): error TS7027: Unreachable code detected.
[08:16:16] TypeScript: 4 semantic errors
[08:16:16] TypeScript: emit succeeded (with errors)
[08:16:17] Finished 'build' after 6.93 s

guillaumejenkins:
Hi @paulpatarinski, thanks a lot for reporting these errors. Could you let us know what version of TypeScript you are using to compile the extension sources?

dlebu:
Fixed by #96. Thanks for trying our extension and please keep the feedback coming.

PR #96, opened 2016-04-07
Fix build errors.

dlebu:
Addresses issues #93 and #95.

### DIFF
--- a/debugger/webkit/webKitDebugAdapter.ts
+++ b/debugger/webkit/webKitDebugAdapter.ts
@@ -345,7 +345,7 @@ export class WebKitDebugAdapter implements IDebugAdapter {
             .map((lineNumber, i) => this._webKitConnection.debugger_setBreakpointByUrl(url, lineNumber, cols ? cols[i] : 0));
 
         // Join all setBreakpoint requests to a single promise
-        return Promise.all(responsePs);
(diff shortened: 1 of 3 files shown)


debugger/webkit/webKitDebugAdapter.ts
```diff
@@ -345,7 +345,7 @@ export class WebKitDebugAdapter implements IDebugAdapter {
             .map((lineNumber, i) => this._webKitConnection.debugger_setBreakpointByUrl(url, lineNumber, cols ? cols[i] : 0));
 
         // Join all setBreakpoint requests to a single promise
-        return Promise.all(responsePs);
+        return <any>Promise.all(responsePs);
```
digeff: Can't we give the proper types to the anys?

dlebu: I checked the code and it looks legit to me. Here's what the documentation says:
> `Promise.all` returns a promise that resolves when all of the promises in the iterable argument have resolved.
I'm not sure why it is not supported in this version of TS, we might hit a bug. I wouldn't want to change working code just for making the compiler happy, so I think leaving it like that is the best option we have right now.

digeff:
:+1:

PR #101, opened 2016-05-05
copy cordova-ionic keyboard plugin

ridomin:
Copy all the typings for ionic projects


rido-min:
I'm talking about the "cordova-ionic/plugins/keyboard" plugin that is used by default in the ionic 1 starter apps.
I think there is a missing file. Without typings\cordova-ionic\index.d.ts we don't have intellisense for the keyboard plugin in app.js

PR #103, opened 2016-05-09
We now parse scripts for source map references instead of relying on the browser

guillaumejenkins:


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -25,10 +26,12 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
     private adbPortForwardingInfo: { targetDevice: string, port: number };
     private ionicLivereloadProcess: child_process.ChildProcess;
     private ionicDevServerUrl: string;
+    private cordovaPathTransforer: CordovaPathTransformer;
```
MSLaguana: Typo: should be `cordovaPathTransformer`, you are missing an `m`

guillaumejenkins: Fixed!

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -654,4 +659,64 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
         // Wait on all the cleanups
         return Q.allSettled([adbPortPromise, killServePromise]).then(() => void 0);
     }
+
+    protected onScriptParsed(script: WebKitProtocol.Debugger.Script): void {
+        if (!script.sourceMapURL && path.extname(script.url) === '.js') {
+            // Browsers don't always report source maps for scripts, so even though no source map was reported for this script, parse it in case it has a sourceMappingUrl attribute.
+            let clientPath = this.cordovaPathTransformer.getClientPath(script.url);
```
MSLaguana: I would prefer it if we could fetch the script contents from the webview somehow rather than trying to work out the file on disk at this point. Not sure how much of a hassle that is though.

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -654,4 +659,64 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
         // Wait on all the cleanups
         return Q.allSettled([adbPortPromise, killServePromise]).then(() => void 0);
     }
+
+    protected onScriptParsed(script: WebKitProtocol.Debugger.Script): void {
+        if (!script.sourceMapURL && path.extname(script.url) === '.js') {
```
MSLaguana: I think this should also check for whether sourcemaps are enabled. We can stash the launch/attach args (have to check in both cases because launching to browser doesn't go through our attach path) and check on them here.

guillaumejenkins: But what about scenarios where the browser reports source maps? `script.sourceMapUrl` will be set. How do we currently enforce `"sourceMaps": false` in that case, and why would that not work in this scenario?

MSLaguana: I believe that the WebkitDebugAdapter (Or maybe just the sourcemap transformer) checks for the `sourceMaps` setting before doing anything with sourcemaps. If it is false, then it does nothing and leaves things unchanged.

guillaumejenkins: Ah, in that case I think it makes sense, so I added the check

src/debugger/cordovaDebugAdapter.ts
guillaumejenkins: We have discussed this offline. In short, there is a way to do this via the Chrome debugging protocol, but performance-wise, it might be better to leave it like this for now. We will revisit in the future if reading from the disk causes issues.

PR #104, opened 2016-05-09
Initial vscode-cordova and cordova-simulate integration

dlebu:
Hey everyone, 

I am starting this PR to get initial feedback on the vscode-cordova and cordova-simulate integration. This change contains:

- A new `Cordova: Simulate` command in the command palette.
- Hosting the simulation host inside VS Code.
- A new debugging target for browser simulation.
- Tests for the command palette integration.

@guillaumejenkins @hamiltonia @panarasi

### DIFF
--- a/package.json
+++ b/package.json
@@ -190,10 +190,21 @@
                         "ionicLiveReload": true
                     },
                     {
-                        "name": "Simulate in browser",
+                        "name": "Simulate Android in browser",
                         "type": "cordova",
                         "request": "launch",
-                        "platform": "simulate",
+                        "platform": "android",
+                        "target": "chrome",
+                        "port": 8000,
+                        "cwd": "${workspaceRoot}"
+                    },
+                    {
+                        "name": "Simulate iOS in browser",
+                        "type": "cordova",
+                        "request": "launch",
+                        "platform": "ios",
+                        "target": "chrome",
+                        "port": 8000,
                         "cwd": "${workspaceRoot}"
                     }
                 ],
--- a/package.json
+++ b/package.json
@@ -210,7 +210,7 @@
                         "properties": {
                             "platform": {
                                 "type": "string",
-                                "description": "The platform to run on"
+                                "description": "The target platform to run for (either 'ios' or 'android'; other platforms are not currently supported)"
                             },
                             "cwd": {
                                 "type": "string",
@@ -263,12 +263,12 @@
                             },
                             "ionicLiveReload": {
                                 "type": "boolean",
-                                "description": "Whether or not to start an Ionic live reload session. This is currently only supported for Ionic 1 projects.",
+                                "description": "Whether or not to start an Ionic live reload session (this is currently only supported for Ionic 1 projects)",
                                 "default": false
                             },
                             "devServerPort": {
                                 "type": "number",
-                                "description": "The port on which Ionic's live reload server should listen on"
+                                "description": "The port on which Ionic's live reload server should listen"
                             },
                             "devServerAddress": {
                                 "type": "string",
--- a/package.json
+++ b/package.json
@@ -225,7 +225,7 @@
                             },
                             "target": {
                                 "type": "string",
-                                "description": "Either 'device', 'emulator', or identifier for a specific device / emulator",
+                                "description": "Either 'device', 'emulator', or identifier for a specific device / emulator. For simulation in the browser, you can use 'chrome'.",
                                 "default": "emulator"
                             },
                             "sourceMaps": {
--- a/package.json
+++ b/package.json
@@ -44,8 +44,12 @@
                 "title": "Cordova: Prepare"
             },
             {
-                "command": "cordova.simulate",
-                "title": "Cordova: Simulate"
+                "command": "cordova.simulate.android",
+                "title": "Cordova: Simulate Android in browser"
+            },
+            {
+                "command": "cordova.simulate.ios",
+                "title": "Cordova: Simulate iOS in browser"
             },
             {
                 "command": "ionic.build",
--- a/package.json
+++ b/package.json
@@ -193,12 +193,8 @@
                         "name": "Simulate in browser",
                         "type": "cordova",
                         "request": "launch",
-                        "platform": "android",
-                        "target": "browser",
-                        "port": 9222,
-                        "sourceMaps": true,
-                        "cwd": "${workspaceRoot}",
-                        "ionicLiveReload": false
+                        "platform": "simulate",
+                        "cwd": "${workspaceRoot}"
                     }
                 ],
                 "configurationAttributes": {
@@ -343,7 +339,7 @@
         "plist-with-patches": "^0.5.1",
         "q": "^1.4.1",
         "source-map": "^0.5.3",
-        "taco-simulate": "dlebu/taco-simulate#dev",
+        "cordova-simulate": "microsoft/cordova-simulate",
         "vscode-extension-telemetry": "0.0.5",
         "winreg": "0.0.13",
         "ws": "0.8.0"
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "bugs": "https://github.com/Microsoft/vscode-cordova/issues",
     "license": "SEE LICENSE IN LICENSE.txt",
     "engines": {
-        "vscode": "^0.10.1"
+        "vscode": "^0.10.7"
     },
     "categories": [
         "Debuggers",
@@ -42,6 +42,10 @@
             {
                 "command": "cordova.prepare",
                 "title": "Cordova: Prepare"
+            },
+            {
+                "command": "cordova.simulate",
+                "title": "Cordova: Simulate"
             },
             {
                 "command": "ionic.build",
@@ -184,6 +188,17 @@
                         "cwd": "${workspaceRoot}",
                         "devServerAddress": "localhost",
                         "ionicLiveReload": true
+                    },
+                    {
+                        "name": "Simulate in browser",
+                        "type": "cordova",
+                        "request": "launch",
+                        "platform": "android",
+                        "target": "browser",
+                        "port": 9222,
+                        "sourceMaps": true,
+                        "cwd": "${workspaceRoot}",
+                        "ionicLiveReload": false
                     }
                 ],
                 "configurationAttributes": {
@@ -320,6 +335,7 @@
     "scripts": {
         "vscode:prepublish": "gulp",
         "compile": "gulp",
+        "postinstall": "node ./node_modules/vscode/bin/install",
         "test": "node ./node_modules/mocha/bin/mocha --recursive -u tdd ./out/debugger/test/ ./out/test/debugger"
     },
     "dependencies": {
@@ -327,6 +343,7 @@
         "plist-with-patches": "^0.5.1",
         "q": "^1.4.1",
         "source-map": "^0.5.3",
+        "taco-simulate": "dlebu/taco-simulate#dev",
         "vscode-extension-telemetry": "0.0.5",
         "winreg": "0.0.13",
         "ws": "0.8.0"
@@ -345,7 +362,7 @@
         "sinon": "^1.17.2",
         "tslint": "^2.5.1",
         "typescript": "^1.6.2",
-        "vscode": "0.10.x",
+        "vscode": "^0.11.0",
         "vsce": "1.0.0"
     }
 }
--- a/src/cordova.ts
+++ b/src/cordova.ts
@@ -92,7 +92,8 @@ export function activate(context: vscode.ExtensionContext): void {
         let ionicTypings: string[] = [
             path.join("angularjs", "angular.d.ts"),
             path.join("jquery", "jquery.d.ts"),
-            path.join("ionic", "ionic.d.ts")
+            path.join("ionic", "ionic.d.ts"),
+            path.join("cordova-ionic", "plugins", "keyboard.d.ts")
         ];
         TsdHelper.installTypings(CordovaProjectHelper.getOrCreateTypingsTargetPath(cordovaProjectRoot), ionicTypings, cordovaProjectRoot);
     }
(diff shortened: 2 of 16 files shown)


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -66,7 +70,7 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
                     });
                 }).then(() => {
                     // For the serve platform, we call super.launch(), which already attaches. For other platforms, attach here
```
MSLaguana: This comment should be updated, perhaps `For the browser platforms, we call super.launch()[...]`?

test/cordova.test.ts
```diff
@@ -24,59 +24,70 @@ suite("VSCode Cordova extension - intellisense and command palette tests", () =>
...
 
-    function checkTypeDefinitions(expectedTypedDefs: string[])
-    {
+    function checkTypeDefinitions(expectedTypedDefs: string[]) {
         let actualTypeDefs = testUtils.enumerateListOfTypeDefinitions(testProjectPath);
         assert.deepEqual(actualTypeDefs, expectedTypedDefs);
     };
 
-    test('Plugin type definitions are installed on activation', () => {
+    test('#Plugin type definitions are installed on activation', () => {
```
MSLaguana: Why have you added the `#`?

typings/taco-simulate/taco-simulate.d.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+declare module "cordova-simulate" {
```
MSLaguana: Should this file be `cordova-simulate.d.ts`?

package.json
```diff
@@ -184,6 +188,13 @@
                         "cwd": "${workspaceRoot}",
                         "devServerAddress": "localhost",
                         "ionicLiveReload": true
+                    },
+                    {
+                        "name": "Simulate in browser",
+                        "type": "cordova",
+                        "request": "launch",
+                        "platform": "simulate",
+                        "cwd": "${workspaceRoot}"
```
guillaumejenkins: I think we should add more properties: `port`, and `targetCordovaPlatform` (android, ios, windows, browser).

Actually, maybe having `platform` and `targetCordovaPlatform` is confusing; is there a way we could keep `platform` to mean "cordova platform", by using another property for `simulate`? What about using the existing `target` property: `"target": "cordova-simulate"`?

package.json
```diff
@@ -345,7 +358,7 @@
         "sinon": "^1.17.2",
         "tslint": "^2.5.1",
         "typescript": "^1.6.2",
-        "vscode": "0.10.x",
+        "vscode": "^0.11.0",
```
guillaumejenkins: I'm just curious, why are we not trying to move to the latest vscode (1.x.x)?

src/extension/simulate.ts
```diff
@@ -0,0 +1,114 @@
...
+        let provider = new SimHostContentProvider(this.simulateInfo.simHostUrl);
+        this.registration = vscode.workspace.registerTextDocumentContentProvider("cordova-simulate", provider);
+        return <any>vscode.commands.executeCommand("vscode.previewHtml", this.simulateUri, vscode.ViewColumn.Two);
+    }
+
+    public launchServer(projectDirectory: string): Q.Promise<simulate.SimulateInfo> {
+        return this.isServerRunning()
+            .then((isRunning: boolean) => {
+                if (!isRunning) {
+                    return simulate.launchServer({ platform: "browser", target: this.target, dir: projectDirectory })
```
guillaumejenkins: Should we let users use other platforms than just browser? Also, I think letting them select a different port is a must, since we cannot assume the default port will always be available on their machine

src/extension/simulate.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public launchAppHost(): Q.Promise<any> {
+        return simulate.launchBrowser(this.target, this.simulateInfo.appUrl);
+    }
+
+    public launchSimHost(): Q.Promise<any> {
+        let provider = new SimHostContentProvider(this.simulateInfo.simHostUrl);
+        this.registration = vscode.workspace.registerTextDocumentContentProvider("cordova-simulate", provider);
+        return <any>vscode.commands.executeCommand("vscode.previewHtml", this.simulateUri, vscode.ViewColumn.Two);
```
guillaumejenkins: What happens if you use `vscode.ViewColumn.Three`? Does it by any chance make sim-host occupy only the last third of the editor, just as if 3 separate documents were open? That could be an interesting solution to making sim-host take less space :)

src/cordova.ts
```diff
@@ -58,13 +59,15 @@ export function activate(context: vscode.ExtensionContext): void {
...
-
     watcher.onDidChange((e: vscode.Uri) => updatePluginTypeDefinitions(cordovaProjectRoot));
     watcher.onDidDelete((e: vscode.Uri) => updatePluginTypeDefinitions(cordovaProjectRoot));
     watcher.onDidCreate((e: vscode.Uri) => updatePluginTypeDefinitions(cordovaProjectRoot));
-
     context.subscriptions.push(watcher);
-    let extensionServer: ExtensionServer = new ExtensionServer();
+
+    let simulator = new PluginSimulator();
+    context.subscriptions.push(simulator);
```
digeff: Is this needed? The extensionServer is already disposing of the pluginSimulator on it's own dipose method...

src/cordova.ts
```diff
@@ -75,6 +78,8 @@ export function activate(context: vscode.ExtensionContext): void {
         () => CordovaCommandHelper.executeCordovaCommand(cordovaProjectRoot, "build")));
     context.subscriptions.push(vscode.commands.registerCommand('cordova.run',
         () => CordovaCommandHelper.executeCordovaCommand(cordovaProjectRoot, "run")));
+    context.subscriptions.push(vscode.commands.registerCommand('cordova.simulate',
```
digeff: You should add this command to the "activationEvents" list. If you want, adding the other missing commands would also be great! :)

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -378,6 +382,21 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
         });
     }
 
+    private launchSimulate(launchArgs: ICordovaLaunchRequestArgs, projectType: IProjectType): Q.Promise<void> {
+        return Q(void 0)
+            .then(() => {
+                let messageSender = new messaging.ExtensionMessageSender();
+                return messageSender.sendMessage(messaging.ExtensionMessage.SIMULATE);
+            }).then((appHostUrl: string) => {
+                launchArgs.url = appHostUrl;
```
digeff: Lines 391-396 are pretty similar to 427-433. Maybe you could extract both of them to a common private method.

test/cordova.test.ts
```diff
@@ -24,59 +24,70 @@ suite("VSCode Cordova extension - intellisense and command palette tests", () =>
...
+    test('#Verify that the commands registered by Cordova extension are loaded', () => {
         return vscode.commands.getCommands(true)
-        .then((results) => {
-            let cordovaCmdsAvailable = results.filter((commandName: string) => {
-                return commandName.indexOf("cordova.") > -1
+            .then((results) => {
+                let cordovaCmdsAvailable = results.filter((commandName: string) => {
+                    return commandName.indexOf("cordova.") > -1
+                });
+                assert.deepEqual(cordovaCmdsAvailable, ["cordova.prepare", "cordova.build", "cordova.run", "cordova.simulate"])
```
digeff: Is the order of cordovaCmdsAvailable deterministic? If not we should sort it before comparing it.

test/cordova.test.ts
```diff
@@ -24,59 +24,70 @@ suite("VSCode Cordova extension - intellisense and command palette tests", () =>
...
+                let androidBuildPath = path.resolve(testProjectPath, "platforms", "android", "build");
+                assert.ok(CordovaProjectHelper.existsSync(androidBuildPath));
+                return testUtils.removeCordovaComponents("platform", testProjectPath, ["android"])
+            });
+    });
+
+    test('#Verify that the simulate command launches the simulate server', () => {
+        return testUtils.addCordovaComponents("platform", testProjectPath, ["browser"])
+            .then(() => vscode.commands.executeCommand("cordova.simulate"))
+            .then(() => Q.delay(10000))
```
digeff: Is there any way to wait for an event instead of waiting always 10 seconds?

test/cordova.test.ts
```diff
@@ -24,59 +24,70 @@ suite("VSCode Cordova extension - intellisense and command palette tests", () =>
...
+            .then(() => testUtils.isUrlReachable('http://localhost:8000/simulator/index.html'))
+            .then((simHostStarted: boolean) => assert.equal(simHostStarted, true, "The simulation host is running."))
```
digeff: You can use assert(simHostStarted, "The simulation host is running.") here.

test/cordova.test.ts
```diff
@@ -24,59 +24,70 @@ suite("VSCode Cordova extension - intellisense and command palette tests", () =>
...
+            .then(() => testUtils.isUrlReachable('http://localhost:8000/index.html'))
+            .then((appHostStarted: boolean) => assert.equal(appHostStarted, true, "The application host is running."))
+            .then(() => testUtils.removeCordovaComponents("platform", testProjectPath, ["browser"]));
```
digeff: I think this line should be a .finally. We want to run it even if the test fails.

src/extension/simulate.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+}
+
+/**
+ * Content provider hosting the simulation UI inside a document.
+ */
+class SimHostContentProvider implements vscode.TextDocumentContentProvider {
+    private simHostUrl: string;
+
+    constructor(simHostUrl) {
```
digeff: This should be a string

src/extension/simulate.ts
```diff
@@ -0,0 +1,114 @@
...
+                }
+            }).then(() => {
+                return this.simulateInfo;
+            });
+    }
+
+    private isServerRunning(): Q.Promise<boolean> {
+        let deferred = Q.defer<boolean>();
+        if (this.simulateInfo) {
+            http.get(this.simulateInfo.simHostUrl, function (res) {
```
digeff: It seems we are turning http.get into a promise in at least 3 places. Maybe you should extract that logic to some utility class.

src/extension/simulate.ts
```diff
@@ -0,0 +1,114 @@
...
+    private simulateUri = vscode.Uri.parse("cordova-simulate://authority/cordova-simulate");
+    private target = "chrome";
+
+    public simulate(projectDirectory: string): Q.Promise<any> {
+        return this.launchServer(projectDirectory)
+            .then(() => this.launchAppHost())
+            .then(() => this.launchSimHost());
+    }
+
+    public launchAppHost(): Q.Promise<any> {
```
digeff: Is the resolution value of the promise intended to be used? If not I think that using void would mark that intent better than any.

digeff:
Looks good :+1: , with some small comments

package.json
dlebu: I think that is a good idea. I'll make the change. How about using the `request` field for simulate, the `platform` for the cordova platform, and `target` for the browser? Currently it is chrome only, but simulate supports multiple browsers.

MSLaguana: If you change the `request` field then you'll have to make some more significant changes to the debug adapter. Right now that field is either `launch` or `attach` and that is handled by code above ours.

typings/taco-simulate/taco-simulate.d.ts
dlebu: Fixed.

package.json
dlebu: I ended up removing the need for a `simulate` platform or request. The decision that simulate needs to happen is now based on the `target`. If the target is one of the supported browsers, we simulate in the browser.

package.json
dlebu: That does not reflect the version of the VSCode IDE, but the `vscode` npm package and `0.11.x` is the latest version.

src/extension/simulate.ts
dlebu: We now have support for all platforms and selecting a custom port. This is for debugging. For the command palette, I need to talk to Parashu to understand how we allow users customizing their simulate experience. My proposal would be through custom settings.

src/extension/simulate.ts
dlebu: It does show in the third column, but only if you already had 2 files open already. This is definitely a possible solution.

src/cordova.ts
dlebu: Fixed. Thanks.

test/cordova.test.ts
dlebu: `deepEqual` ignores order to my knowledge so we don't need to sort.

test/cordova.test.ts
dlebu: Fixed, thanks!

test/cordova.test.ts
dlebu: Not sure, I just followed the recipe of the existing tests.

test/cordova.test.ts
dlebu: Fixed.

src/extension/simulate.ts
dlebu: Indeed.

test/cordova.test.ts
dlebu: It separates the test name from the suite name in test results. With a suite name so long, it is easier to follow.

src/cordova.ts
dlebu: I'm not sure if we want that. I will follow up with @panarasi .

src/extension/simulate.ts
dlebu: Fixed.

test/cordova.test.ts
dlebu: Removed.

dlebu:
Thanks everyone, merging to master.

PR #107, opened 2016-05-12
Updated 3rd party dependencies.

dlebu:
Also, added OSSREADME.json and npm-shrinkwrap.json. The setup is similar to what VSCode has for their repository.


dlebu:
Recalling this PR as I will submit a new one against master.

PR #110, opened 2016-05-19
Adding telemetry for cordova-simulate

guillaumejenkins:
- Added more telemetry properties to the launch event for cordova-simulate
- Created a telemetry wrapper compatible with cordova-simulate's telemetry events in order to send cordova-simulate telemetry

### DIFF
--- a/package.json
+++ b/package.json
@@ -200,6 +200,7 @@
                         "platform": "android",
                         "target": "chrome",
                         "port": 8000,
+                        "livereload": true,
                         "cwd": "${workspaceRoot}"
                     },
                     {
@@ -209,6 +210,7 @@
                         "platform": "ios",
                         "target": "chrome",
                         "port": 8000,
+                        "livereload": true,
                         "cwd": "${workspaceRoot}"
                     }
                 ],
@@ -284,6 +286,14 @@
                             "devServerAddress": {
                                 "type": "string",
                                 "description": "The IP address that Ionic should use for the live reload server"
+                            },
+                            "livereload": {
+                                "type": "boolean",
+                                "description": "When simulating in the browser, determines whether live reload is enabled"
+                            },
+                            "forceprepare": {
+                                "type": "boolean",
+                                "description": "When simulating in the browser, determines whether a file change triggers a cordova prepare before live reloading"
                             }
                         }
                     },
--- a/src/utils/cordovaSimulateTelemetry.ts
+++ b/src/utils/cordovaSimulateTelemetry.ts
@@ -1,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for details.
+
+import {Telemetry} from './telemetry';
+import {TelemetryGenerator, TelemetryHelper} from './telemetryHelper';
+
+/**
+ * This class is a telemetry wrapper compatible with cordova-simulate's telemetry. Cordova-simulate expects an object with a sendTelemetry() function, and calls it every time there is a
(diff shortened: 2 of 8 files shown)


src/utils/cordovaSimulateTelemetry.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import {Telemetry} from './telemetry';
+import {TelemetryGenerator, TelemetryHelper} from './telemetryHelper';
+
+/**
+ * This class is a telemetry wrapper compatible with cordova-simulate's telemetry. Cordova-simulate expects an object with a sendTelemetry() function, and calls it every time there is a
+ * telemetry event. This wrapper creates a telemetry event compatible with vscode-cordova's telemetry, based on cordova-simulate's event, and sends it using the Telemetry module.
+ */
+export class CordovaSimulateTelemetry {
+    public sendTelemetry(eventName: string, props: Telemetry.ITelemetryProperties, piiProps: Telemetry.ITelemetryProperties): void {
```
dlebu: Are all the parameters required? Do we need parameter validation here?

dlebu:
Minor comment, this looks good to me.

package.json
```diff
@@ -284,6 +286,14 @@
                             "devServerAddress": {
                                 "type": "string",
                                 "description": "The IP address that Ionic should use for the live reload server"
+                            },
+                            "livereload": {
+                                "type": "boolean",
+                                "description": "When simulating in the browser, determines whether live reload is enabled"
+                            },
+                            "forceprepare": {
```
digeff: Should we use camel case here?

guillaumejenkins: I thought about that but cordova-simulate uses all lowercase options... We could always use camel case in launch.json, and then translate that to lowercase for cordova-simulate consumption, but I just felt it was better to leave the same options that cordova-simulate uses

src/utils/cordovaSimulateTelemetry.ts
guillaumejenkins: I guess we could say that cordova-simulate's "telemetry API" is strict, in the sense that it always calls `sendTelemetry` with these 3 arguments, with proper default values if they were empty or undefined.

I can add some quick checks to validate the arguments on top of that, as it wouldn't hurt, but I'm inclined to leave the parameters non-optional

src/utils/cordovaSimulateTelemetry.ts
```diff
@@ -0,0 +1,28 @@
...
+        let fullEventName = 'cordova-simulate-' + eventName;
```
digeff: Is there any kind of convention for events name? I think that in taco we used to do TACO/build

src/utils/cordovaSimulateTelemetry.ts
```diff
@@ -0,0 +1,28 @@
...
+        let generator = new TelemetryGenerator(fullEventName);
+        let telemetryEvent: Telemetry.TelemetryEvent = new Telemetry.TelemetryEvent(fullEventName);
```
digeff: You don't need to type this variable. You can also use const if you are not re-assigning it.

dlebu: There is no side effect of this approach other than type clarity. Why is this an issue?

digeff: I think that adding types when they add clarity is a good think. Personally, in this case, doing
let someVar: MyClass = new MyClass(...) doesn't give me any extra clarity because I can already see new MyClass next to it so I already know the type. I personally find it easier to read const someVar = new MyClass(...), but if you guys feel this is clearer I'm fine with it.

digeff:
:+1: with some minor comments

PR #116, opened 2016-05-26
Fixing build error.

dlebu:
@panarasi @guillaumejenkins @MSLaguana


MSLaguana:
Looks good to me. I thought that the `.bin` path would work too, but this is probably more resilient.

Issue #112, opened 2016-05-20
Simulate: rename port option to simulatePort


dlebu:
The `port` debugging option is intended to be used for the chrome debugging port. We need to add a new option for the simulation port.

PR #119, opened 2016-05-27
Fixing #112: Simulate: rename port option to simulatePort

dlebu:
#112

### DIFF
--- a/package.json
+++ b/package.json
@@ -199,7 +199,7 @@
                         "request": "launch",
                         "platform": "android",
                         "target": "chrome",
-                        "port": 8000,
+                        "simulatePort": 8000,
                         "livereload": true,
                         "cwd": "${workspaceRoot}"
                     },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -199,7 +199,7 @@
                         "request": "launch",
                         "platform": "android",
                         "target": "chrome",
-                        "port": 8000,
+                        "simulatePort": 8000,
```
MSLaguana: Should we also keep a definition of `port` as well, to show that both can apply?

dlebu: We could, for consistency with the other tasks.

MSLaguana:
Looks good to me otherwise.

package.json
dlebu: We ended up not adding it for the browser scenarios, for consistency with ionic serve scenario, which does not have it defined.

dlebu:
Thanks for the feedback, merging to master!

PR #127, opened 2016-06-02
Added project path hash suffix to the extension server and text docum…

dlebu:
…ent content provider for allowing multiple instances at the same time.

### DIFF
--- a/src/debugger/cordovaDebugAdapter.ts
+++ b/src/debugger/cordovaDebugAdapter.ts
@@ -21,6 +21,7 @@ import {WebKitDebugAdapter} from '../../debugger/webkit/webKitDebugAdapter';
 import {CordovaProjectHelper} from '../utils/cordovaProjectHelper';
 import {IProjectType, ISimulateTelemetryProperties, TelemetryHelper, TelemetryGenerator} from '../utils/telemetryHelper';
 import {settingsHome} from '../utils/settingsHelper';
+import {Telemetry} from '../utils/telemetry';
 
 export class CordovaDebugAdapter extends WebKitDebugAdapter {
     private static CHROME_DATA_DIR = 'chrome_sandbox_dir'; // The directory to use for the sandboxed Chrome instance that gets launched to debug the app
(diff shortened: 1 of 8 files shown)


src/debugger/debugCordova.ts
```diff
@@ -9,9 +9,10 @@ import {DebugSession} from '../../debugger/common/debugSession';
 
 import {Telemetry} from '../utils/telemetry';
 
-let version = JSON.parse(fs.readFileSync(path.join(__dirname, '..', '..', '..', 'package.json'), 'utf-8')).version;
+let projectRoot = path.join(__dirname, '..', '..', '..');
+let version = JSON.parse(fs.readFileSync(path.join(projectRoot, 'package.json'), 'utf-8')).version;
 
 // Enable telemetry, forced on for now.
-Telemetry.init('cordova-tools-debug-adapter', version, {isExtensionProcess: false}).then(() => {
+Telemetry.init('cordova-tools-debug-adapter', version, {isExtensionProcess: false}, projectRoot).then(() => {
```
MSLaguana: I believe this is passing in the wrong projectRoot. It refers to `__dirname` which will the the directory containing the source of our extension, not the directory containing the user's project which is what the IPC is based on now.

src/utils/telemetry.ts
```diff
@@ -91,10 +91,10 @@ export module Telemetry {
             isExtensionProcess: boolean;
         }
 
-        export function init(appNameValue: string, appVersion: string, initOptions: ITelemetryInitOptions): Q.Promise<any> {
+        export function init(appNameValue: string, appVersion: string, initOptions: ITelemetryInitOptions, projectRoot: string): Q.Promise<any> {
```
MSLaguana: Perhaps make the projectRoot part of the ITelemetryInitOptions, since it is only used in some cases?

MSLaguana: Actually, I'd suggest replacing `isExtensionProcess` with something like `masterProjectRoot`, since the project root is required exactly when this is not the (in-process) server.

src/debugger/debugCordova.ts
MSLaguana: I don't know if we have a good way of finding the project location at this stage, the first information the debug adapter would get is when the launch/attach request comes in, but that's too late for a bunch of error scenarios that we'd like to find out about...

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -830,6 +835,21 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
...
+    private initializeTelemetry(projectRoot: string): Q.Promise<any> {
+        if (!this.telemetryInitialized) {
+            this.telemetryInitialized = true;
+            let version = JSON.parse(fs.readFileSync(path.join(__dirname, '..', '..', '..', 'package.json'), 'utf-8')).version;
+            // Enable telemetry, forced on for now.
+            return Telemetry.init('cordova-tools-debug-adapter', version, { isExtensionProcess: false, projectRoot: projectRoot })
+                .catch((e) => {
+                    this.outputLogger('Could not initialize telemetry.' + e.message || e.error || e.data || e);
+                });
+        }
```
MSLaguana: This else case must return a promise

MSLaguana:
Looks good apart from the issue with `initTelemetry` in the debug adapter.

dlebu:
Thanks @MSLaguana! Merging to master.

PR #128, opened 2016-06-02
Remove touch mode when resizing viewport

guillaumejenkins:
Setting touch-mode emulation to true prevented clicking the "Resume" button in Chrome when the app was paused at a breakpoint. With this, we no longer set the simulation to touch mode when in a mobile viewport.


MSLaguana:
Looks good to me.

dlebu:
LGTM

PR #130, opened 2016-06-07
Licensing

dlebu:
Summary of changes:

1. Updated the top level ThirdPartyNotices.txt to include only dependencies in the repository.
2. Added a new `release` directory, containing the release third party notices, license, and readme modifications for release. All the license file handling is done by a new `gulp release` task.


gulpfile.js
```diff
@@ -13,17 +14,28 @@ var Q = require('q');
...
     var commandToExecute = cordovaCmd + " " + command;
-    var process = child_process.exec(commandToExecute, { cwd: cwd });
+    return executeCommand(cwd, commandToExecute);
+}
+
+function executeCommand(cwd, commandToExecute) {
+    var deferred = Q.defer();
+    var process = child_process.exec(commandToExecute, { cwd: cwd }, (error, stdout, stderr) => {
+        if (error) {
+            console.error(console.error("An error occurred: " + error));
```
guillaumejenkins: nested `console.error`? Is this on purpose?

ThirdPartyNotices.txt
```diff
@@ -1,607 +1,147 @@
...
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+=========================================
+END OF DefinitelyTyped NOTICES, INFORMATION, AND LICENSE
+
+%% Ionic Documentation NOTICES, INFORMATION, AND LICENSE BEGIN HERE
+=========================================
+Ionic Documentation (http://ionicframework.com/docs/)
+
+� 2013-2016 Drifty Co
```
guillaumejenkins: Some of these characters keep getting changed to this "unknown" character, we should keep an eye out and change them back to their original characters (here this would be a Copyright character)

The other file (`release3party.txt`) could potentially have those as well, but the diff is not shown here, so we should double check in some other way that we don't have such characters

MSLaguana: Garbled character, should be a copyright symbol?

dlebu: @guillaumejenkins @MSLaguana it is a copyright symbol, I will fix it in the next commit.

gulpfile.js
dlebu: not on purpose, thanks!

npm-shrinkwrap.json
```diff
@@ -0,0 +1,1394 @@
+{
```
MSLaguana: Should this be checked in, or just generated on-demand?

dlebu: I would keep it checked in.

MSLaguana:
Looks good to me pending an answer to how the package.json license field is handled.

dlebu:
I will merge this in as is and modify it tomorrow if needed. Thanks for the review!

PR #131, opened 2016-06-08
Restart the server with every simulate call

dlebu:
During bashing I found an issue where the incorrect platform was being served due to reusing the simulate server. Restarting the server with every simulate call will ensure the right platform is being simulated. This PR depends on https://github.com/Microsoft/cordova-simulate/pull/52 being merged.


dlebu:
@guillaumejenkins We need to stop the server in the case of command palette scenarios as well, which don't go through the debugging code path.

PR #133, opened 2016-06-08
Change release task to remove dev dependencies.

dlebu:
`vsce` is not smart enough to remove the dev dependencies at package time, so we need to do this manually. However, we quickly run into a chicken and egg problem if we rely on the locally installed `vsce` to package the extension. To work around this:

- we use `npm prune --production` to remove all dev dependencies before packaging
- we use the globally installed `vsce` for creating the release build

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -125,7 +125,7 @@ gulp.task('watch-build-test', ['build', 'build-test'], function() {
     return gulp.watch(sources, ['build', 'build-test']);
 });
 
-gulp.task('release', function () {
+gulp.task('release', ['build'], function () {
     var licenseFiles = ["LICENSE.txt", "ThirdPartyNotices.txt"];
     var backupFolder = path.resolve(path.join(os.tmpdir(), 'vscode-cordova'));
     if (!fs.existsSync(backupFolder)) {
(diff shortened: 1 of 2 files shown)


gulpfile.js
```diff
@@ -144,9 +144,12 @@ gulp.task('release', function () {
...
             fs.writeFileSync('LICENSE.txt', fs.readFileSync('release/releaselicense.txt'));
             fs.writeFileSync('ThirdPartyNotices.txt', fs.readFileSync('release/release3party.txt'));
-
+        }).then(()=>{
+            console.log("Removing dev dependencies...");
+            return executeCommand(path.resolve(__dirname), 'npm prune --production');
+        }).then(()=>{
             console.log("Creating release package...");
-            return executeCommand(path.resolve(__dirname), 'node ./node_modules/.bin/vsce package');
+            return executeCommand(path.resolve(__dirname), 'vsce package');
```
dlebu: Since `vsce` is a dev dependency and will be removed by `npm prune --production`, I relied on the globally installed one here.

dlebu: If you guys have a better idea on how to automate this, please let me know!

MSLaguana:
Are you sure about this? I just tried it out and `npm ls | wc` gives me ~750 lines of output in the project, while in the packaged vsix I get from `gulp release` I get ~260 lines of output which means it is removing a bunch of stuff.

MSLaguana:
Ah, I see that you are right and `vsce@1.0.0` does not correctly handle npm 3 style node_modules folders, and `vsce@1.5.0` can't handle this project either since there's some npm error with the cordova-simulate package.

dlebu:
I'll merge this in and once we publish to npm I'll try the other solution we talked about.

dlebu:
Thanks for reviewing @MSLaguana!

Issue #134, opened 2016-06-09
Debug configuration "Serve to the browser (Ionic serve)" timeouts


sandcake:
Launching the Debugger with the configuration "Serve to the browser (Ionic serve)" fails with the following message in the console:
"Starting the Ionic dev server timed out (10000 ms)",
although manually launching an "ionic serve" call succeeds.

Finally VSCode shows the following error at the command palette:
"[cordova-debug-adapter] Error: Starting the Ionic dev server timed out (10000 ms)"

Using cordova 6.2.0 and Ionic 1.7.15

sandcake:
Not knowing where to start for contributing, i would simply tell how i fixed it on my end.

Increasing the _serverReadyTimeout_ variable in the function _startIonicDevServer_ in **src/debugger/cordovaDebugAdapter.ts** from 10000 to 30000 (the unit here is ms) helped.

`private startIonicDevServer(launchArgs: ICordovaLaunchRequestArgs, cliArgs: string[]): Q.Promise<string> {
        //code related to other things ....

        let serverReadyTimeout: number = 30000; //previously 10000

        //further code doing some other things where the variable we edited is actually used.
        
}`

dlebu:
Hi @sandcake, thanks for the feedback. Can you give us a little more details about your environment? We are interested in the following:

- version of VSCode
- operating system for your dev machine
- assuming you are the same person that reviewed our extension in the VSCode Marketplace (wizardsteve22), can you tell us more about the failures you had while trying the other debug configurations?
- can you confirm that cordova/ionic works on your machine when used from the command line, without using our tools?

Thanks,
Daniel.

guillaumejenkins:
@sandcake I would add another question to dlebu's: when you run `ionic serve` from the command line, how long does it take for you to start seeing output from the Ionic dev server? Something like the following:

```
Running live reload server: ...
Watching: ...
Running dev server: ...
```

I'm just trying to understand if our tool could be slowing down the launch of the Ionic server on your machine.

sandcake:
Hi @dlebu to my environment:

- VSCode version: 1.2.0
- OS of dev machine:Windows 8.1
- The assumption is correct. I just went through the issue list, and after updating cordova and ionic the other debugging configurations (other from Run serve) now fail for understandable reasons, such as "no devices attached" for the "Run Android on device" configuration. And i don't have any way to produce the old error cases since i don't know which versions of cordova and ionic i previously had.
[...]

PR #138, opened 2016-06-09
Exposed Ionic server timeout as a launch.json property

guillaumejenkins:
Addresses #134

### DIFF
--- a/package.json
+++ b/package.json
@@ -287,6 +287,11 @@
                                 "type": "string",
                                 "description": "The IP address that Ionic should use for the live reload server"
                             },
+                            "devServerTimeout": {
+                                "type": "number",
+                                "description": "Timeout for starting the Ionic dev server when serving to the browser or running with Ionic live reload enabled.",
+                                "default": 10000
+                            },
                             "simulatePort": {
                                 "type": "number",
                                 "description": "Port to use for connecting to the local Cordova Simulate server.",
--- a/package.json
+++ b/package.json
@@ -271,7 +271,7 @@
                             },
                             "appStepLaunchTimeout": {
                                 "type": "number",
-                                "description": "Timeout for individual steps when launching iOS app",
+                                "description": "Timeout in milliseconds for individual steps when launching iOS app",
                                 "default": 5000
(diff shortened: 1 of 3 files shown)


MSLaguana:
Looks good to me.

package.json
```diff
@@ -287,6 +287,11 @@
                                 "type": "string",
                                 "description": "The IP address that Ionic should use for the live reload server"
                             },
+                            "devServerTimeout": {
+                                "type": "number",
+                                "description": "Timeout for starting the Ionic dev server when serving to the browser or running with Ionic live reload enabled.",
```
dlebu: We should mention the unit as well - 

> Timeout, in milliseconds, for starting...

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -596,7 +596,7 @@ export class CordovaDebugAdapter extends WebKitDebugAdapter {
         let errorRegex: RegExp = /error:.*/i;
         let serverReady: boolean = false;
         let appReady: boolean = false;
-        let serverReadyTimeout: number = 10000;
+        let serverReadyTimeout: number = launchArgs.devServerTimeout || 10000;
```
dlebu: Would we want to bump up the default as well?

AlGantori:
I just started using the cordova tools extension for ionic2,  and I see time out (20000 ms) where exactly in the launch.json I need to make the change.  Please indicate the syntax to bump this default up. 
Do I need to also install the "Chrome extension" ? Thank you.

AlGantori:
To whoever stumbles here
OK, I browsed the "Files changed" above and guessed, I added
, "devServerTimeout" : 30000
and this time it worked and launched on Chrome to my amazement (Firefox is my default browser)
Do I need to also install the "Chrome extension" ? Thank you.

MSLaguana:
You don't need the chrome extension installed unless you want to manually launch/attach chrome. Our extension uses the chrome debugging extension internally to support Cordova scenarios. We don't support debugging firefox since the debugging protocols are different.

Issue #141, opened 2016-06-13
typings for contacts plugin is empty


rido-min:
I noticed that when I add the contacts plugin, the typings file added to the project is an empty file named contact.d.ts when this should be contacts.d.ts
https://github.com/Microsoft/vscode-cordova/blob/master/pluginTypings.json#L12

PR #142, opened 2016-06-13
fix: typo on contacts typings

ridomin:
Adding the contacts plugin adds an empty Contact.d.ts file. Looks like a typo (missing plural) in the mapping file.

This could close https://github.com/Microsoft/vscode-cordova/issues/141


dlebu:
This looks good to me. Thanks!

PR #155, opened 2016-06-27
Added InAppBrowser member to Cordova interface.

### LINKED ISSUE #154: Typings for InAppBrowser seem off
I'm having difficulty getting the intellisense for InAppBrowser to work right. 

1) The docs (https://www.npmjs.com/package/cordova-plugin-inappbrowser) show that InAppBrowser is available via the cordova object, but if you type cordova VSC will not suggest InAppBrowswer:

image

2) If I type "InAp..." then VSC does suggest it, which implies it is in the global scope, but it isn't. 

3) If I'm at the end of InAppBrowser, VSC doesn't seem to recognize it as it doesn't suggest open:

image

dlebu:
Fixes #154

### DIFF
--- a/CordovaTypings/cordova/plugins/InAppBrowser.d.ts
+++ b/CordovaTypings/cordova/plugins/InAppBrowser.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Apache Cordova InAppBrowser plugin.
+// Type definitions for Apache Cordova InAppBrowser plugin 1.4.0.
 // Project: https://github.com/apache/cordova-plugin-inappbrowser
 // Definitions by: Microsoft Open Technologies, Inc. <http://msopentech.com>
 // Definitions: https://github.com/borisyankov/DefinitelyTyped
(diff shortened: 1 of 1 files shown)


CordovaTypings/cordova/plugins/InAppBrowser.d.ts
```diff
@@ -2,7 +2,7 @@
 // Project: https://github.com/apache/cordova-plugin-inappbrowser
 // Definitions by: Microsoft Open Technologies, Inc. <http://msopentech.com>
 // Definitions: https://github.com/borisyankov/DefinitelyTyped
-//
```
MSLaguana: Perhaps add a comment here referencing the version number this corresponds to?

MSLaguana:
Looks good to me.

CordovaTypings/cordova/plugins/InAppBrowser.d.ts
dlebu: Done. Thanks!

Issue #156, opened 2016-06-28
Running on Ubuntu 16


raducrisan1:
Hello,
I have two issues using the extension from Ubuntu 16.04. I tried on different machines and the behavior is the same all the time. 

I wrote down the steps I followed to set up the Cordova app:

I have a fresh install of Ubuntu 16.04, which has node 4.2.6
I installed cordova globally from bash: `sudo npm install -g cordova` . The installed version is 6.2.0
I created a first app from bash: `cordova create simpleapp`
From simpleapp directory I added the android platform: `cordova platform add android`. The installed version is 5.1.1
I installed Android Studio 2.1.2 to help me have the emulator and the Android SDK installed. Here is the path I added into the ~/.bashrc file: `PATH=$PATH:/home/radu/Android/Sdk/platform-tools:/home/radu/Android/Sdk/tools`. I also added the ANDROID_HOME environment variable. The output of `echo $ANDROID_HOME` is: `/home/radu/Android/Sdk`
I created a single Android Virtual Device based on Marshmallow X64 system image. Then I opened the simpleapp folder from VS Code

The first issue is related to build. I pressed F6 to build the app and the message coming from the extension is: `/bin/sh: 1: cordova.cmd: not found`, although I am on Ubuntu 16.04.
![buildcordova](https://cloud.githubusercontent.com/assets/15193363/16430110/a0bc50c8-3d82-11e6-91c5-7120b54ac005.png)

The second issue is related to debugging: I couldn't properly use the debug mode, VS Code did not attach to the emulator. In the debug mode I selected Run Android on Emulator and then I pressed F5 to start debugging. The emulator started, the app opened and after the message `Attaching to android` I encountered another message: `Unable to find emulator`. 
![debugcordova](https://cloud.githubusercontent.com/assets/15193363/16430252/6558371c-3d83-11e6-943c-5c1520866eb6.png)

The emulator had the simple app in front, like below:
![emulator](https://cloud.githubusercontent.com/assets/15193363/16430339/d6ae034c-3d83-11e6-9853-e8811dcf2384.png)

Please let me know what to do in order to have the issues fixed. Thank you!

MSLaguana:
Thanks for reporting these issues. We haven't done a lot of testing outside of windows and mac, and it looks like we've got some platform-dependent bits in there that don't account for linux.

[...]

Issue #100, opened 2016-04-23
Debug fails with [cordova-debug-adapter] Error: Unable to find device


porjo:
Attempting to do `Run Android on device` results in this message in console:
```
Launching for android (This may take a while)...
App successfully launched
Attaching to android
Unable to find device
```
However the 'HelloCordova' app is successfully running on my device! (and was launched by the above action)

I'm using VSCode on Fedora Linux 23. I have Android Studio installed on the same system and it works with my device just fine.

MSLaguana:
What is the exact output of `adb devices`, and please include the line where you input the command. That error message indicates that we tried to find a target listed as a `device`, but we were unsuccessful. We have not tested extensively on linux, and if there is a change in the output format that may cause us issues.

porjo:
```
$ adb devices
List of devices attached
0123456789ABCDEF	device
```

MSLaguana:
To be sure, if you try to launch on android via VS Code again it still gives you the original error that you posted? It looks like it should work from what you show there.

porjo:
I've tried again from a freshly booted system, and the same thing happened again - error in VSCode console, yet app gets launched on my device.

MSLaguana:
Could you please do the following to help us diagnose where we are going wrong:

* Open up your project in VS code with our extension, and try to build, getting the initial error
* Open up a console, and run `adb devices > /tmp/devices.out 2> /tmp/devices.err` to get the exact, whitespace-preserving output of adb devices
* Upload the /tmp/devices.out and devices.err files somewhere. Feel free to anonymise the device ID if you wish, but just replace the device ID with Xs or something, be sure not to change the whitespace. You can e-mail it to me at jithomso@microsoft.com if you don't want to put it somewhere public

Hopefully that will let me see what we are missing and work out a fix.

porjo:
Here is the output file. The `.err` file was empty so I've left that off.

[devices.txt](https://github.com/Microsoft/vscode-cordova/files/235562/devices.txt)

MSLaguana:
The regex that we use to detect devices works fine on that file, so I'm not sure why it is failing for you. I'll have to get a linux machine and test it our myself to see what's going on.

porjo:
FYI, an interesting observation - I tried running vscode under strace e.g.: `strace -o process_dump -f code` and found that it the issue does NOT occur when run that way. Instead, console shows this message instead:
```
Launching for android (This may take a while)...
App successfully launched
Attaching to android
[...]

PR #158, opened 2016-06-30
Fix spawn timing on linux.

dlebu:
Fixing #156 and #100 .

### DIFF
--- a/src/debugger/extension.ts
+++ b/src/debugger/extension.ts
@@ -21,7 +21,7 @@ export function execCommand(command: string, args: string[], errorLogger: (messa
     proc.on('error', (err: Error) => {
         deferred.reject(err);
     });
-    proc.on('exit', (code: number) => {
+    proc.on('close', (code: number) => {
         if (code !== 0) {
             errorLogger(stderr);
             errorLogger(stdout);


src/debugger/extension.ts
```diff
@@ -21,7 +21,7 @@ export function execCommand(command: string, args: string[], errorLogger: (messa
     proc.on('error', (err: Error) => {
         deferred.reject(err);
     });
-    proc.on('exit', (code: number) => {
+    proc.on('close', (code: number) => {
```
MSLaguana: Does close have a code? Perhaps we should still have an exit handler to reject on code != 0, and have the execCommand function return `Q.all([processExited, processClosedOutput]).then(() => stdout)`

dlebu: `close` does have a code. For our scenario, we are safe to use `close` in a similar fashion to `exit` because we are not inheriting any of the parent process streams. https://nodejs.org/api/child_process.html#child_process_event_close

MSLaguana:
Looks good to me.

dlebu:
@MSLaguana @guillaumejenkins Thanks! Merging to master.

PR #163, opened 2016-07-06
fix error extending Window

ridomin:
This will fix the error: TS2430: Build:Interface 'InAppBrowser' incorrectly extends interface 'Window'


MSLaguana:
Looks like this is a similar fix to https://github.com/DefinitelyTyped/DefinitelyTyped/commit/20762bfa1c3bbb08d17e66b2826ac73ed3085f82

I'd suggest also making the changes to the comment as they do there, or more specifically taking the latest version from definitelytyped which also has a couple of other minor changes.

ridomin:
Including changes in comments, however I there is still a delta, since the DT version does not include this https://github.com/Microsoft/vscode-cordova/commit/4b3fa677d3c9505388e686a7f3ea22ce372cb50f

MSLaguana:
Looks good to me. Thanks Rido

dlebu:
Thank you @ridomin!

PR #166, opened 2016-07-07
Refactor chromeDebugAdapter to use node package

PatoBeltran:
Instead of having the source code for the chrome debug adapter use the node package. This way we can have the latest version of the package and manage it correctly as a dependency.

We are still using some parts that are not part of the public API, so this makes us depend on changes they do to their private API.


package.json
```diff
@@ -1,4 +1,5 @@
 {
+<<<<<<< 124e84b9751524558d5d5335a954014fd409ee2d
```
geof90: Seems like an unresolved merge conflict.

PatoBeltran: Fixed 👍

MSLaguana:
After I found out about https://github.com/Microsoft/vscode-cordova/pull/172 I would like to wait until vscode-chrome-debug-core also updates their `ws` dependency to avoid npm install failing, hopefully they'll oblige soon.

PatoBeltran:
@MSLaguana I just updated the version of `vscode-chrome-debug-core`. It now has a fix for the typing issues and it has `ws` updated.

PatoBeltran:
@guillaumejenkins talked with LCA and changed third party notices!

Issue #168, opened 2016-07-12
Option to disable XHR Proxy


ChiroxV:
I didn't find an option to turn the xhr proxy off. With the xhr proxy I get the following issue:
_Failed to load resource: net::ERR_EMPTY_RESPONSE (http://localhost:8000/xhr_proxy?rurl=https://companyservice/)_

After commenting out the content of the init()-function in xhr-proxy.js it worked perfectly.

corodva-simulate documented a property named corsproxy to turn it off. But I can't set it in the debugging options (launch.json) or simulator options (when running without debugger). Adding this option to a future release would be great.

MSLaguana:
I'd like to try and understand why the XHR proxy doesn't work for you, are you able to share a sample of how you are trying to use it?

I will also expose the ability to disable the proxy, that's quite reasonable.

ChiroxV:
I will try to reproduce it in a smaller environment to create a sample. If possible...

MSLaguana:
Were you able to get a smaller example of when the XHR proxy failed you @ChiroxV ?

ChiroxV:
Sorry for my late response. No, I can only reproduce it in the company environment. I assume it's a proxy or firewall setting. @MSLaguana

panarasi:
@ChiroxV Thanks for reporting the issue. We tried reproducing the issue but did not have any luck. We are also not hearing this issue from other customers. 
Since we do not have a repro, lets close this issue. If you think that this is a problem, can you please re-open it ?

sijad:
I have similar problem

```
$ curl 'http://localhost:8000/xhr_proxy?rurl=http%3A//x.x.x.x/path/login' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36' -H 'Content-Type: text/plain' --data 'grant_type=password&username=usr&password=pwd' -v
*   Trying ::1...
* Connected to localhost (::1) port 8000 (#0)
> POST /xhr_proxy?rurl=http%3A//x.x.x.x/path/login HTTP/1.1
> Host: localhost:8000
> Accept: */*
> User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36
> Content-Type: text/plain
[...]

PR #169, opened 2016-07-12
Adding the ability to disable XHR proxying

MSLaguana:
As requested in #168, this exposes the ability to disable XHR proxying. Note that disabling this option will mean that XHR requests often fail for CORS or CSP reasons unless you put in your own workarounds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -309,6 +309,11 @@
                                 "type": "string",
                                 "description": "The directory where temporary browser simulation files are hosted.",
                                 "default": "${workspaceRoot}/.vscode/simulation"
+                            },
+                            "corsproxy": {
+                                "type": "boolean",
+                                "description": "When simulating in the browser, determines whether XHR requests are proxied to appear as though they originate from the same domain as the target.",
+                                "default": true
                             }
                         }
                     },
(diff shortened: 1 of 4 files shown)


MSLaguana:
@Microsoft/cordova-tools

toddm:
:shipit:

mbraude:
Are you just exposing this to customers? It doesn't look like it actually does the work.

MSLaguana:
Yes, this is just exposing the ability to opt-out of the XHR proxying that cordova-simulate does. The option is parsed [here](https://github.com/Microsoft/cordova-simulate/blob/aee0215a0d67f177d865476f595c148e3d482055/src/server/simulator.js#L217) which then determines whether [this](https://github.com/Microsoft/cordova-simulate/blob/aee0215a0d67f177d865476f595c148e3d482055/src/server/socket.js#L198) starts up the proxy, which is implemented [here](https://github.com/Microsoft/cordova-simulate/blob/master/src/modules/xhr-proxy.js)

Issue #168 states that our proxy failed but without the proxy the request worked.

mbraude:
Ah I see. Makes sense & looks good to me.

PR #171, opened 2016-07-13
Removing ion_list from ion_refresh snippet

### LINKED ISSUE #102: ionic html snippet ion_refresh shouldn't include ion-list
The ionic html snippet ion_refresher snippet includes an ion-list and not just ionic refresher code.  This feels unneeded since there is an ion_listSimple and ion_listComplex.  As well I normally end up having to delete the ion list since by the time I am ready to implement the refresher I already have a working ion list.

MSLaguana:
The ion_refresh example used a list, but a list is not inherently required in order to use ion_refresh.

Fixes #102


MSLaguana:
@Microsoft/cordova-tools

toddm:
:shipit:

PR #172, opened 2016-07-14
Changing ws package to avoid install failure

MSLaguana:
ws depends on bufferutils, and that fails to install on windows typically. The dependency is optional from a functionality perspective, but marked as required in ws@0.8.0 and thus it fails the install. Updating to 1.0.1 (which is the same as used by another dependency) fixes the issue.

This should get our CI builds working again.

@Microsoft/cordova-tools


PatoBeltran:
LGTM!

MSLaguana:
Unfortunately our CI system is still attempting to install these packages, even when I build from this branch. Locally things seem to be fine, but I must be missing something else.

MSLaguana:
With some minor changes to the CI build, this branch now successfully builds. Waiting to merge it until we update the third party notices.

toddm:
:shipit:

PR #173, opened 2016-07-26
Add support for ionic 2

PatoBeltran:
Add support for scenarios when using ionic 2

### DIFF
--- a/src/cordova.ts
+++ b/src/cordova.ts
@@ -82,7 +82,7 @@ export function activate(context: vscode.ExtensionContext): void {
                     generator.add("visibleTextEditorsCount", vscode.window.visibleTextEditors.length, false);
                 });
         }).then(() => {
-            return simulator.simulate(options);
+            return simulator.simulate(options, projectType);
         });
     };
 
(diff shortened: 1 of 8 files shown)


PatoBeltran:
@Microsoft/cordova-tools please CR 😄

src/cordova.ts
```diff
@@ -107,11 +107,15 @@ export function activate(context: vscode.ExtensionContext): void {
     // Install Ionic type definitions if necessary
     if (CordovaProjectHelper.isIonicProject(cordovaProjectRoot)) {
         let ionicTypings: string[] = [
-            path.join("angularjs", "angular.d.ts"),
             path.join("jquery", "jquery.d.ts"),
-            path.join("ionic", "ionic.d.ts"),
```
MSLaguana: Is this because we defer to the typescript sources for ionic 2 rather than external typings?

src/extension/simulate.ts
```diff
@@ -8,6 +8,7 @@ import * as cordovaServer from "cordova-serve";
 import * as path from "path";
 import * as simulate from "cordova-simulate";
 import {CordovaSimulateTelemetry} from "../utils/cordovaSimulateTelemetry";
+import {IProjectType} from '../utils/telemetryHelper';
```
MSLaguana: Since the `IProjectType` is now more closely associated with product behavior, I'd suggest moving it from telemetryHelper into another file more indicative of it's usage. Otherwise I'd be inclined to think that it is just an additional nice-to-know thing.

src/utils/telemetryHelper.ts
```diff
@@ -150,7 +150,7 @@ export class TelemetryGenerator extends TelemetryGeneratorBase {
 }
 
 export interface IProjectType {
-    ionic: boolean;
+    ionic1: boolean;
```
MSLaguana: Might want to run this change past @panarasi since it may require changes to the people consuming the telemetry.

MSLaguana:
Looks like the support for ionic2 was already there, this is just improving detection and removing some blocks, is that right?

src/cordova.ts
PatoBeltran: This is correct

PatoBeltran:
@MSLaguana that's right. Most of the support was enabled by https://github.com/Microsoft/vscode-cordova/pull/166. This PR just takes care of knowing correctly when a project is ionic 2 and some small bug fixes.

src/utils/telemetryHelper.ts
panarasi: I think we should leave Ionic 1 as ionic - as it used to be, and use ionic2 as a new string. This way, we don't mess up existing events.

PatoBeltran:
@MSLaguana Addressed CR comments

MSLaguana:
Looks good to me

PR #176, opened 2016-08-11
Searching ANDROID_HOME for adb

### LINKED ISSUE #149: Android: use ANDROID_HOME if adb and android aren't in the PATH
Right now, for Android-related actions, we rely on `adb` and `android` being in the PATH. Instead of failing when these are not found, we should at least look if ANDROID_HOME is defined, and use that to reach the executables.

Most setup guides for Android SDK / Cordova only mention setting ANDROID_HOME (and don't mention PATH at all).

MSLaguana:
With this change, we match cordova functionality and succeed if adb is not on the PATH, but is in ANDROID_HOME/platform-tools.

Fixes #149


TimBarham:
:shipit:

PR #177, opened 2016-08-12
Take newer cordova simulate

MSLaguana:
Updating the typings and interactions to use the newer interfaces in cordova-simulate.

Not ready to merge yet, need to wait for cordova-simulate to be published so we take the correct npm dependency and verify nothing else changed there.


BretJohnson:
@MSLaguana I pushed my changes.  Want to take a look, esp for the shrinkwrap changes. Thx.

MSLaguana:
Looks good to me, although it seems that I did this branch before checking in the updated shrinkwrap for the currently published build which is causing a conflict. Hopefully a simple "take mine" merge for that file should work out, and the diff should be smaller than it is now.

MSLaguana:
Looks good.

Issue #58, opened 2016-03-10
[cordova-debug-adapter] TypeError: Cannot read property 'nodeValue' of undefined


Ritzlgrmft:
In between, I am able to debug my app in the simulator (#54). But now I wanted to debug on an attached iPhone. When I select `Run iOS on device`, I see in the debug console:

    Launching for ios (This may take a while)...
    Launching app (This may take a while)...
    Installing app on device
    Launching app

And the I get the error

    [cordova-debug-adapter] TypeError: Cannot read property 'nodeValue' of undefined

When I select `Attach to running iOS on device`, I see in the debug console:

    Attaching to ios
    Configuring debugging proxy

And then I get the same error as above.

MSLaguana:
I recall seeing an issue like this when there was a misbehaving plist file... Do you have any plugins in this project? We had a bug in TACO which was triggered by an extension containing a plist file, so it may be the same problem here.

MSLaguana:
Do you have a `platforms/ios/<projectname>/<projectname>-Info.plist` file? If so, what happens if you try running `/usr/libexec/PlistBuddy -c Print path/to/<projectname>-Info.plist`?

Ritzlgrmft:
The output from PlistBuddy is:

    Dict {
        CFBundleName = ${PRODUCT_NAME}
        CFBundleIdentifier = markus.recipes.mobile
        CFBundleInfoDictionaryVersion = 6.0
        CFBundleIcons = Dict {
        }
        CFBundleVersion = 1.1.0
        NSMainNibFile~ipad = 
        CFBundleExecutable = ${EXECUTABLE_NAME}
        LSRequiresIPhoneOS = true
        UIRequiresFullScreen = true
        CFBundleIcons~ipad = Dict {
        }
        CFBundleDisplayName = ${PRODUCT_NAME}
        NSAppTransportSecurity = Dict {
            NSAllowsArbitraryLoads = true
        }
        CFBundlePackageType = APPL
        CFBundleIconFile = icon.png
        CFBundleSignature = ????
        NSMainNibFile = 
        CFBundleDevelopmentRegion = English
        CFBundleShortVersionString = 1.1.0
    }

Any yes, I am using some plugins:

    <plugin name="cordova-plugin-app-version" spec="~0.1.8" />
    <plugin name="cordova-plugin-device" spec="~1.1.1" />
    <plugin name="cordova-plugin-dialogs" version="1.2.0" />
[...]

PR #185, opened 2016-08-29
Fix issue #58 - Replace plist patches with plist.

jicongw:
This is to resolve issue #58 
Replace plist-with-patches with plist.
Update test and shrinkwrap file.


MSLaguana:
Looks good to me

jicongw:
Thanks @MSLaguana :)

PR #197, opened 2016-10-18
Add support for crosswalk webview (for android)

nicolashenry:
I had an error "Cannot connect to the target: read ECONNRESET" since I use the crosswalk webview (https://crosswalk-project.org/documentation/cordova.html) because the localabstract name is not "webview_devtools_remote_{pid}" but "{app.package}_devtools_remote".

Here is what I did :
- I changed the `ps | grep ${package}` to a simple 'ps' with a post filtering because grep can match a wrong line.
For example, if I have two apps "my.app" and "my.app.sub", grep could return "my.app.sub" for "ps | grep my.app".
- I retrieved the sockets inodes linked to the pid with `ls -l /proc/${pid}/fd`
- I retrieved the localabstract name linking inodes with 'cat /proc/net/unix' filtered inspired by : https://cs.chromium.org/chromium/src/chrome/browser/devtools/device/android_device_info_query.cc?l=144

This seems to work but there may be a better way. (Note that I don't know if the crosswalk webview works with IOS)

### DIFF
--- a/src/debugger/cordovaDebugAdapter.ts
+++ b/src/debugger/cordovaDebugAdapter.ts
@@ -300,6 +300,8 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
                             if (fields.length < 8) {
                                 continue;
                             }
+                            // flag = 00010000 (16) -> accepting connection
+                            // state = 01 (1) -> unconnected
                             if (fields[3] !== '00010000' || fields[5] !== '01') {
                                 continue;
                             }
--- a/src/debugger/cordovaDebugAdapter.ts
+++ b/src/debugger/cordovaDebugAdapter.ts
@@ -262,14 +262,67 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
                 return parsedFile.attrib[packageKey];
             });
         return Q.all([packagePromise, adbDevicesResult]).spread((appPackageName, targetDevice) => {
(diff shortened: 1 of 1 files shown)


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -262,14 +262,67 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
...
+                .then(socketsInodes => !socketsInodes || socketsInodes.length === 0 ? undefined :
+                    this.runAdbCommand(getSocketsCommandArguments, errorLogger)
+                    .then((getSocketsResult) => {
+                        const lines = getSocketsResult.split('\n');
+                        for (const line of lines) {
+                            const fields = line.split(/[ \r]/);
+                            if (fields.length < 8) {
+                                continue;
+                            }
+                            if (fields[3] !== '00010000' || fields[5] !== '01') {
```
MSLaguana: Please document what these checks are actually looking for.

MSLaguana:
This looks good to me! Could you please just add a little bit of documentation to clarify some of the magic numbers?

MSLaguana:
Thanks for your contribution!

To answer your concern about iOS, attaching for debugging works very differently in iOS. Either crosswalk exposes itself in the same way that safari webviews do, or we won't be able to connect to it at all right now.

nicolashenry:
@MSLaguana done in 303f79e

PR #211, opened 2016-11-30
Allow VSCode to provide typings when possible

vladimir-kotikov:
VSCode starting from 1.7.2 automatically provides typings for dependencies mentioned in package.json (in Cordova projects this usually happens when plugin has been installed using `--fetch` option) - in this case we can safely ignore own typings installation mechanism and rely on VSCode. Previous behavior is still retained for previous VSCode versions though.

This PR also disables installation of typings into Ionic 2 projects, because they are written in typescript and Ionic provides own set of TS wrappers around plugins.

### DIFF
--- a/src/cordova.ts
+++ b/src/cordova.ts
@@ -113,10 +113,10 @@ export function activate(context: vscode.ExtensionContext): void {
             path.join("cordova-ionic", "plugins", "keyboard.d.ts")
         ];
         if (CordovaProjectHelper.isIonic1Project(cordovaProjectRoot)) {
-            ionicTypings.push[
+            ionicTypings = ionicTypings.concat([
                 path.join("angularjs", "angular.d.ts"),
                 path.join("ionic", "ionic.d.ts")
-            ];
+            ]);
         }
         TsdHelper.installTypings(CordovaProjectHelper.getOrCreateTypingsTargetPath(cordovaProjectRoot), ionicTypings, cordovaProjectRoot);
     }
--- a/src/cordova.ts
+++ b/src/cordova.ts
@@ -10,6 +10,7 @@ import {CordovaProjectHelper} from './utils/cordovaProjectHelper';
 import {CordovaCommandHelper} from './utils/cordovaCommandHelper';
 import {ExtensionServer} from './extension/extensionServer';
 import * as Q from "q";
+import * as semver from 'semver';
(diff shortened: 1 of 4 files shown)


src/cordova.ts
```diff
@@ -212,7 +213,31 @@ function getRelativeTypeDefinitionFilePath(projectRoot: string, parentPath: stri
...
+        // Read installed node modules and filter out plugins that have been already installed in node_modules
+        // This happens if user has used '--fetch' option to install plugin. In this case VSCode will provide
+        // own intellisense for these plugins using ATA (automatic typings acquisition)
+        try {
+            const installedNpmModules: string[] = fs.readdirSync(nodeModulesDir);
+            installedPlugins = installedPlugins
+                .filter(pluginId => {
+                    // plugins with `forceInstallTypings` flag don't have typings on NPM yet,
+                    // so we still need to install these even if they present in 'node_modules'
+                    return getPluginTypingsJson()[pluginId].forceInstallTypings ||
```
MSLaguana: I think you might need to check whether `getPluginTypingsJson()[pluginId]` exists before accessing `forceInstallTypings`, otherwise if there is a plugin installed that we do not explicitly cater for this will throw an exception.

src/cordova.ts
```diff
@@ -212,7 +213,31 @@ function getRelativeTypeDefinitionFilePath(projectRoot: string, parentPath: stri
 }
 
 function updatePluginTypeDefinitions(cordovaProjectRoot: string): void {
+    // We don't need to install typings for Ionic2 since it has own TS
+    // wrapper around core plugins
+    if (CordovaProjectHelper.isIonic2Project(cordovaProjectRoot)) {
```
MSLaguana: Might want to also check for the existence of a `tsconfig.json`, and avoid installing typings if one is found.

MSLaguana:
Looks good apart from those 2 comments.

vladimir-kotikov:
@MSLaguana, addressed, thanks!

vladimir-kotikov:
Added a logic for persisting and maintaining references to installed type definitions in `cordova-plugins.d.ts` file.

vladimir-kotikov:
@MSLaguana, could you take a look once more?

vladimir-kotikov:
Thanks! 👍

PR #227, opened 2017-01-05
Upgrading vscode-chrome-debug-core version

### LINKED ISSUE #201: Debug console error Unimplemented console API: undefined
`Unimplemented console API: undefined`

hello, 

I am getting this error message instead of any exception messages or console.log messages when i try to debug ionic cordova project on an android device.  App is working fine on device.

do you have any ideas how to solve it?

thank you.

MSLaguana:
Taking the latest version of vscode-chrome-debug-core to fix issues with newer versions of chrome. Some architecture changes have happened, so there were some tweaks required including a bit of a hack to get access to a path remapper.

Fixes #201


gulpfile.js
```diff
@@ -58,7 +58,7 @@ var sources = [
 
 var projectConfig = {
     noImplicitAny: false,
-    target: 'ES5',
+    target: 'es6',
```
toddm: Does case matter for this?  You've got it as ES6 near the bottom in another change.

MSLaguana: I don't think it matters, since both directly invoking `tsc` and using the gulp build seem to compile the code fine. I'll change it to be consistent though.

Issue #213, opened 2016-12-05
VSCode keeps failing with error "dev server timed out" on "Serve to the browser (Ionic serve)" timeout  


AlGantori:
ionic RC0 main upgrade steps seems to have taking me to RC3, which is great :)
**CLI ionic server works fine on the newly created v2 project.**
Yesterday, before the upgrade, while under 2.0.0.beta11 VSCode debugging was working fine launching a separate window of chrome.
BUT now, VSCode fails debugging into the browser with this notorious timeout error

I even tried 420000 no difference
launch.json....
```
{
            "name": "Serve to the browser (ionic serve)",
            "type": "cordova",
            "request": "launch",
            "platform": "serve",
            "cwd": "${workspaceRoot}",
            "devServerAddress": "localhost",
            "sourceMaps": true,
            "ionicLiveReload": true
            , "devServerTimeout" : 420000
            , "port": 9222
            
        },

```
here is my package.json
```
 "dependencies": {
    "@angular/common": "2.1.1",
    "@angular/compiler": "2.1.1",
    "@angular/compiler-cli": "2.1.1",
    "@angular/core": "2.1.1",
    "@angular/forms": "2.1.1",
    "@angular/http": "2.1.1",
    "@angular/platform-browser": "2.1.1",
    "@angular/platform-browser-dynamic": "2.1.1",
    "@angular/platform-server": "2.1.1",
    "@ionic/app-scripts": "0.0.46",
    "@ionic/storage": "1.1.6",
    "ionic-angular": "2.0.0-rc.3",
    "ionic-native": "2.2.3",
    "ionicons": "3.0.0",
    "rxjs": "5.0.0-beta.12",
    "zone.js": "0.6.26"
  },
  "devDependencies": {
    "@ionic/app-scripts": "0.0.46",
    "typescript": "^2.0.10"
  },

```
ionic info 
```
Your system information:

 ordova CLI: 6.4.0
Ionic Framework Version: 2.0.0-rc.3
Ionic CLI Version: 2.1.14
Ionic App Lib Version: 2.1.7
Ionic App Scripts Version: 0.0.46
ios-deploy version: Not installed
ios-sim version: Not installed
OS: Windows 7
Node Version: v7.2.0
Xcode version: Not installed

```
npm -v
`4.0.3
`
Windows 7 64bit.

"Run on Android device" works though a bit slow

I have also run 
 **cordova platform add browser**
Adding browser project...
[...]

PR #230, opened 2017-01-10
Make possible to serve to the browser for Ionic2 projects

matrosov-nikita:
- Check logs from Ionic2 dev server to determine whether it's ready

- Increase default server timeout since loading server for Ionic2 projects takes longer

- Use 'tree-kill' to kill properly processes spawned by 'ionic serve'

There was bug with finding an open port for the notification server.It's strongly recommended to update 'ionic-app-scripts' to the 0.0.48 version otherwise you get an error: 'Error: listen EADDRINUSE 0.0.0.0:53703'

Related issues: https://github.com/Microsoft/vscode-cordova/issues/134, https://github.com/Microsoft/vscode-cordova/issues/213


src/debugger/extension.ts
```diff
@@ -69,21 +70,7 @@ export function cordovaStartCommand(args: string[], cordovaRootPath: string): ch
...
-        ];
-
-        return execCommand(command, args, errorLogger).then(() => void 0);
-    } else {
-        childProcess.kill();
-
-        return Q<void>(void 0);
-    }
+export function killChildProcess(childProcess: child_process.ChildProcess): Q.Promise<void> {
+    TreeKill(childProcess.pid, 'SIGKILL');
```
MSLaguana: Is this change necessary, or just convenient? Where possible we've tried to limit the number of third party dependencies we have.

matrosov-nikita: @MSLaguana, I think this change is necessary. There are two processes spawned by serve command - [first](https://github.com/Microsoft/vscode-cordova/blob/034090b374c2ed8b00e5e1616fd374c1d8e54788/src/debugger/extension.ts#L69) is spawned by extension and [second](https://github.com/driftyco/ionic-cli/blob/master/lib/utils/npmScripts.js#L72) is spawned by ionic-cli. The second process remains in memory after debug session is being terminated.  Using `tree-kill` package is the only solution which I could find to kill that second process.

MSLaguana: I see, fair enough.

PR #237, opened 2017-01-31
Adding telemetry to report all plugins in the project

toddm:
Adding telemetry to report all plugins in the project during activation and then to report any plugins added after that during launch and attach.

### DIFF
--- a/src/utils/cordovaProjectHelper.ts
+++ b/src/utils/cordovaProjectHelper.ts
@@ -28,6 +28,38 @@ export class CordovaProjectHelper {
     private static IONIC_PROJECT_FILE: string = "ionic.project";
     private static IONIC_LIB_DEFAULT_PATH: string = path.join("www", "lib", "ionic");
 
+    private static CORE_PLUGIN_LIST: string[] = ["cordova-plugin-battery-status",
+                                                "cordova-plugin-camera",
+                                                "cordova-plugin-console",
+                                                "cordova-plugin-contacts",
+                                                "cordova-plugin-device",
+                                                "cordova-plugin-device-motion",
+                                                "cordova-plugin-device-orientation",
+                                                "cordova-plugin-dialogs",
+                                                "cordova-plugin-file",
+                                                "cordova-plugin-file-transfer",
+                                                "cordova-plugin-geolocation",
+                                                "cordova-plugin-globalization",
+                                                "cordova-plugin-inappbrowser",
+                                                "cordova-plugin-media",
+                                                "cordova-plugin-media-capture",
+                                                "cordova-plugin-network-information",
+                                                "cordova-plugin-splashscreen",
+                                                "cordova-plugin-statusbar",
+                                                "cordova-plugin-vibration",
+                                                "cordova-plugin-ms-azure-mobile-apps",
+                                                "cordova-plugin-hockeyapp",
+                                                "cordova-plugin-code-push",
+                                                "cordova-plugin-bluetoothle",
+                                                "phonegap-plugin-push",
+                                                "cordova-plugin-ms-azure-mobile-engagement",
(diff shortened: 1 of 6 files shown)


src/utils/cordovaProjectHelper.ts
```diff
@@ -115,6 +147,27 @@ export class CordovaProjectHelper {
...
     }
 
+    public static getInstalledPluginDetails(projectRoot: string, pluginId: string): any {
+        let packageJsonPath: string = path.resolve(projectRoot, CordovaProjectHelper.PROJECT_PLUGINS_DIR, pluginId, 'package.json');
+
+        if (!CordovaProjectHelper.existsSync(packageJsonPath)) {
+            return null;
+        }
+
+        let packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'));
```
MSLaguana: I would suggest moving this into the try, just in case there is malformed JSON somehow

src/utils/cordovaProjectHelper.ts
```diff
@@ -115,6 +147,27 @@ export class CordovaProjectHelper {
...
+        let details: any = new Object();
```
MSLaguana: Could you please create an interface type describing the expected fields for this object? I would also be inclined to change to something along the lines of 

```
let details: IPluginDetails = {
    PluginId: packageJson.name,
    Version: packageJson.version,
    PluginType: CordovaProjectHelper.CORE_PLUGIN_LIST.indexOf(pluginId) >= 0 ? 'Core' : 'Npm'
};
return details
```

rather than creating an arbitrary object and then immediately filling in the fields.

src/utils/telemetryHelper.ts
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for details.
 
 import {CordovaProjectHelper} from './cordovaProjectHelper';
+import {IPluginDetails} from './cordovaProjectHelper';
```
MSLaguana: You can combine this with the previous import like `import {CordovaProjectHelper, IPluginDetails} from './cordovaProjectHelper';`

src/utils/telemetryHelper.ts
```diff
@@ -239,6 +240,58 @@ export class TelemetryHelper {
...
+        });
+
+        // If none, return
+        if (newPlugins.length == 0) {
+            return;
+        }
+
+        var pluginDetails : Object[] = new Array<Object>();
+        // Send telemetry event with list of new plugins
+        newPlugins.forEach(pluginName => {
```
MSLaguana: I believe you should be able to do

```
let pluginDetails: IPluginDetails[] =
    newPlugins.map(pluginName => CordovaProjectHelper.getInstalledPluginDetails(projectRoot, pluginName))
    .filter(detail => !!detail);
```

to convert each plugin to its IPluginDetail (or null) and then filter out the nulls.

src/utils/telemetryHelper.ts
```diff
@@ -239,6 +240,58 @@ export class TelemetryHelper {
...
+            let pluginDetail : IPluginDetails = CordovaProjectHelper.getInstalledPluginDetails(projectRoot, pluginName);
+            if (pluginDetail) {
+                pluginDetails.push(pluginDetail);
+            }
+        });
+
+        let pluginEvent = new Telemetry.TelemetryEvent('plugins', { plugins: JSON.stringify(newPlugins) });
```
toddm: I just noticed that this should stringify pluginDetails instead of newPlugins.  I changed it from a simple array to the detailed list but was still sending the simple array.

MSLaguana:
Looks good to me

MSLaguana:
Actually, I keep forgetting: could you please also update the CHANGELOG.md

Issue #236, opened 2017-01-29
Unable to debug ionic 2 app remotely on my android device


sgadekar81:
I am trying to debug ionic app using cordova tools plugin, I have tried reinstalling plugin for several times, on reinstalling first time it works and gives following error since i am new, can any one help?

![image](https://cloud.githubusercontent.com/assets/14721916/22403645/03b755fa-e644-11e6-9daa-0bf0bc2aa2e5.png)

![screenshot_2017-01-29-16-58-35](https://cloud.githubusercontent.com/assets/14721916/22403694/4d921b78-e645-11e6-9257-58077c1f45c8.png)

jmesa-sistel:
add
```
"config": {
    "ionic_bundler": "webpack",
    "ionic_source_map": "source-map"
  },
```
in package.json to be able to debug properly 
Then execute in console: `ionic run android -l`
When the app is running in the device, then click to debug in VSCode using the option:
`Attach to running android on device`
Now you can debug properly and livereload is working, there is only one bug about breakpoints: [issue 235](https://github.com/Microsoft/vscode-cordova/issues/235)

sgadekar81:
thanks, now i am able to debug, thanks for suggestion

matrosov-nikita:
@sgadekar81, what version of Node are you running on? I was able to reproduce this bug using node 6.3.0, but after upgrading to 6.7.0 the problem has disappeared.

sgadekar81:
node v7.0.0;
npm 3.10.8

vladimir-kotikov:
This seems to has been resolved, so closing this. @sgadekar81, feel free to reopen if you're still having any issues. Thanks

Issue #229, opened 2017-01-09
"Run Android on device" runs into endless loop


Cedware:
Hello.
If I choose the debug option "Run Android on device", the plugin seems to get stuck in a loop.
The terminal prints "Launching for android (This may take a while)...", after a while the terminal relaunches with the same message and each time a "strange" log-file appears in the project root with the following filename:
`npm-debug.log.{a-number}`
Those log file have no content.
If i launch the application with:
`cordova run android`
or
`ionic run android`
I can use "attach to running android device" and everything works fine.

MSLaguana:
Can you please copy the exact output of the debugging console when this is happening at some point? There is one time that we attempt to npm install something, and I'd like to see if that is what's happening here.

HugoMarteau:
![image](https://cloud.githubusercontent.com/assets/8122998/22365360/5150dcba-e477-11e6-8916-1bdffea30567.png)

The debbug message in the console. nothing more or less.
Just disapearing and appearing

Cedware:
I'm sorry, I totally thought, I had answered your question.
But HugoMarteau hit the point, the Console doesn't show anything beside the "Launching for android" Text.
If you ne more information, let me know

vladimir-kotikov:
The fix has been merged into master branch and will be available in the next release. Closing this fo now.

PR #239, opened 2017-02-08
Fix running android on device with livereload

matrosov-nikita:
We should look for 'launch success' string instead of 'Ionic server commands' to resolve app deffered, since a logic that prints ionic command tips after launch application was removed in ionic-cli 2.1.9. Also verified that this patch is continuing to work with ionic pre-2.1.9
Related issues: https://github.com/Microsoft/vscode-cordova/issues/236, https://github.com/Microsoft/vscode-cordova/issues/229


MSLaguana:
Looks good, thanks!

Issue #250, opened 2017-03-07
'Ionic: Build' and 'Ionic: Run' commands should allow user to specify platform


vladimir-kotikov:
Curently all contributed commands are just invoking corresponding Cordova/Ionic command without giving use a choice which platform to build/run for. This may lead to odd problems, e.g.:
1. when there are multiple platforms added to the project, running them all (as "Ionic/Cordova: Run" behaves currently) ususally doesn't make sense
1. when iOS platform is added to the app and user tries to call `Ionic: Build` on Windows machine this leads to an error “You cannot run iOS unless you are on Mac OSX”

To choose platform, there should probably be a dropdown, similar to command palette, that would allow to choose platform to run/build. Also it'd be nice to have an ability to save the default value or pick it from workspace configuration.

iTOYS:
Implemented in #328

PR #260, opened 2017-03-14
Use Android as default platform for Ionic on Windows

vladimir-kotikov:
Previous approach was to rely on Ionic, however it uses 'ios' as default one even on windows, which leads to odd “You cannot run iOS unless you are on Mac OSX” error. Instead we'll detect if we're running on Windows and use Android in that case.

This is a quick fix to mitigate the problem, described in #250

### DIFF
--- a/src/utils/cordovaCommandHelper.ts
+++ b/src/utils/cordovaCommandHelper.ts
@@ -32,6 +32,14 @@ export class CordovaCommandHelper {
             generator.add('command', command, false);
             let outputChannel = window.createOutputChannel("cordova");
             let commandToExecute = cliCommandName + " " + command;
+
+            if (useIonic && os.platform() === 'win32' && ['build', 'run'].indexOf(command) >= 0) {
+                // ionic build/run commands use 'ios' as default platform, even on Windows, which
+                // causes them to fail w/ odd “You cannot run iOS unless you are on Mac OSX” error.
+                // To prevent this we enforce these commands to use 'android' in such case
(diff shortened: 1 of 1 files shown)


vladimir-kotikov:
@matrosov-nikita, @alsorokin, could you take a look please?

src/utils/cordovaCommandHelper.ts
```diff
@@ -32,6 +32,14 @@ export class CordovaCommandHelper {
             generator.add('command', command, false);
             let outputChannel = window.createOutputChannel("cordova");
             let commandToExecute = cliCommandName + " " + command;
+
+            if (useIonic && os.platform() === 'win32' && command in ['build', 'run']) {
```
matrosov-nikita: `in` operator uses properties  keys for searching rather than values

src/utils/cordovaCommandHelper.ts
```diff
@@ -32,6 +32,14 @@ export class CordovaCommandHelper {
...
+                // ionic build/run commands use 'ios' as default platform, even on Windows, which
+                // causes them to fail w/ odd “You cannot run iOS unless you are on Mac OSX” error.
+                // To prevent this we enforce these commands to use 'android' in such case
+                commandToExecute += 'android';
```
matrosov-nikita: It seems that space is missing before platform's name

vladimir-kotikov:
Updated, Thanks

PR #283, opened 2017-06-17
Fix to not prepend 'cordova' argument when running ionic serve.

harleybl:
The previous change adds a 'cordova' argument to all debugging command lines. This works for all cases except when running ionic serve (simulate in a browser). That fails when you attempt to run with the current code base. This change addresses that. 

I looked at https://ionicframework.com/docs/cli/ to see if any other debugging CLI commands would make sense to not include the 'cordova' argument, but it seems like serve is the only one.


jcyh0120:
how can I applied this pull request before it can be updated in vscode extension market?

vladimir-kotikov:
@harleybl, thanks

Issue #297, opened 2017-08-07
1.2.9 Version - Debbuger error in VSCode


asierf:
1.2.8 Version was working fine in windows 7.

ERROR : 
[cordova-tools] Error processing "launch": 'Debugger.setAsyncCallStackDepth' wasn't found

Console : 
'Debugger.setAsyncCallStackDepth' wasn't foundError processing "launch": 'Debugger.setAsyncCallStackDepth' wasn't found

iTOYS:
@asierf, thanks for reporting. Is it cordova or ionic app? 
Could you providing precise steps to reproduce problem(which cordova version using, which configuration did you run, etc.)?
Also, could you please try creating an empty cordova project and try debugging it?

asierf:
It's an Ionic v1 app running under emulated android sdk version 19

Ionic version 3.6.0
Cordova version 7.0.1

"name": "Run Android on emulator",
"type": "cordova",
"request": "launch",
"platform": "android",
"target": "emulator",
 "port": 9222,
 "sourceMaps": true,
  "cwd": "${workspaceRoot}",
  "ionicLiveReload": false

I have the same issue debbuging an empty cordova project

vladimir-kotikov:
Ok, so this happens due to bug in vscode-chrome-debug-core, where `Debugger.setAsyncCallStackDepth` method is called during `attach` and while it is [marked as optional](https://github.com/Microsoft/vscode-chrome-debug-core/blob/012a56f8637a380c2117b6fecd096e5412925d08/crdp/crdp.d.ts#L1308) its existense is [not being checked](https://github.com/Microsoft/vscode-chrome-debug-core/blob/50b5fffe4ebc3edb7bf87cc483eb6ec758d361d5/src/chrome/chromeDebugAdapter.ts#L342) before calling. The workaround for now is to catch the exception appropriately and silently continue

vladimir-kotikov:
@asierf this has been fixed in #298 and will be available in the next release that likely will be available EOW.

iTOYS:
#https://github.com/Microsoft/vscode-chrome-debug-core/issues/226

kirajhpang:
hi, I having this issue on debug ionic v2 when 
`Run iOS on simulator `
`Run iOS on device`
`Attach to running iOS on device/simulator`

Debug Console output message show as following:
```
Attaching to ios
Configuring debugging proxy
Attaching to app.
DEVICE READY FIRED AFTER 2742 ms
[...]

PR #298, opened 2017-08-08
Catching Debugger.setAsyncCallStackDepth error

iTOYS:
Fix for #297

### DIFF
--- a/src/debugger/cordovaDebugAdapter.ts
+++ b/src/debugger/cordovaDebugAdapter.ts
@@ -220,6 +220,15 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
                     this.outputLogger('Attaching to app.');
                     this.outputLogger('', true); // Send blank message on stderr to include a divider between prelude and app starting
                     return super.attach(processedAttachArgs)
+                        .catch((err) => {
+                            if (err.message && err.message.indexOf(MISSING_API_ERROR) > -1) {
+                                // Bug in `vscode-chrome-debug-core` calling unimplemented method Debugger.setAsyncCallStackDepth
+                                // just ignore it
+                                // https://github.com/Microsoft/vscode-cordova/issues/297
+                                return void 0;
+                            }
+                            throw err;
+                        })
                         .then(() => {
                             // Safari remote inspector protocol requires setBreakpointsActive
                             // method to be called for breakpoints to work (see #193 and #247)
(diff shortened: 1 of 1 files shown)


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -226,6 +228,12 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
                                 .catch(err => this.outputLogger('Failed to call "setBreakpointsActive" of debugging frontend. Debugging will continue...'));
 
                             this.attachedDeferred.resolve(void 0);
+                        })
+                        .catch((err) => {
```
vladimir-kotikov: IMO `catch` needs to be attached before `then` to catch exteption as early as possible

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -226,6 +228,12 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
                                 .catch(err => this.outputLogger('Failed to call "setBreakpointsActive" of debugging frontend. Debugging will continue...'));
 
                             this.attachedDeferred.resolve(void 0);
+                        })
+                        .catch((err) => {
+                            if (err.message && err.message.indexOf(MISSING_API_ERROR) > -1) {
+                                return this.attachedDeferred.resolve(void 0);
+                            }
+                            return err;
```
vladimir-kotikov: In case if this is not `MISSING_API_ERROR` it's not sufficient to just return error - you need to rethrow it to reject the promise chain

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -1085,9 +1093,16 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
                 this.terminateSession(errMsg);
             });
 
-            return this.doAttach(port, launchUrl, args.address).then(() => {
-                this.attachedDeferred.resolve(void 0);
-            });
+            return this.doAttach(port, launchUrl, args.address)
+                .then(() => {
```
vladimir-kotikov: 2 comments above are also applicable here
- [ ] move `catch` higher i npromise chain
- [ ] rethrow error instead of just returning it

vladimir-kotikov:
As a general note - @iTOYS please also add comment explaining why this workaround has been introduced to be able to identify this afterwards easily

src/debugger/cordovaDebugAdapter.ts
iTOYS: We won't catch error early if just put `catch` before `then`
If an error occurs, we will immediately get the function `catch`

src/debugger/cordovaDebugAdapter.ts
iTOYS: 1. See https://github.com/Microsoft/vscode-cordova/pull/298/files#r131962894
2. done

Issue #302, opened 2017-08-16
Debugging Ionic app using "Run iOS on simulator" fails


vladimir-kotikov:
Repro steps: 
1. Ionic start foo super && cd foo
2. Open directory in VSCode
3. Create default launch.json for Cordova
4. Start debugging with "Run iOS on simulator" configuration

Debug console output:
```
Launching for ios (This may take a while)...
Launching app (This may take a while)...
[GThe emulate command has been renamed. To find out more, run:

  ionic cordova emulate --help

Attaching to ios
Configuring debugging proxy
ENOENT: no such file or directory, scandir '/Users/kotikov.vladimir/sandbox/foo/platforms/ios/build/emulator'
```

This happens because the extension runs 'ionic emulate ios' instead of 'ionic cordova emulate ios'. Looks like [this check](https://github.com/Microsoft/vscode-cordova/blob/master/src/debugger/extension.ts#L84) doesn't work anymore with ionic-cli 3.8.1 and greater.

`Ionic info`: 
```
cli packages: (/Users/kotikov.vladimir/.nvm/versions/node/v8.2.1/lib/node_modules)

    @ionic/cli-utils  : 1.8.1
    ionic (Ionic CLI) : 3.8.1

global packages:

    Cordova CLI : 7.0.1

local packages:

    @ionic/app-scripts : 2.1.3
    Cordova Platforms  : ios 4.4.0
    Ionic Framework    : ionic-angular 3.6.0

System:

    Android SDK Tools : 26.0.2
    ios-deploy        : 1.9.1
    ios-sim           : 6.0.0
    Node              : v8.2.1
    npm               : 5.3.0
    OS                : macOS Sierra
    Xcode             : Xcode 8.2.1 Build version 8C1002
```

vladimir-kotikov:
Fixed in #303

PR #303, opened 2017-08-16
Ionic CLI >=3.8.0 support

iTOYS:
Fix for #302

### DIFF
--- a/src/debugger/extension.ts
+++ b/src/debugger/extension.ts
@@ -83,7 +83,9 @@ export function cordovaStartCommand(args: string[], cordovaRootPath: string): ch
 
     if (cliName === 'ionic' && !isIonicServe) {
         try {
-            let ionicInfo = child_process.spawnSync(command, ['-v']);
+            let ionicInfo = child_process.spawnSync(command, ['-v'], {
+                cwd: cordovaRootPath
+            });
             let ionicVersion = ionicInfo.stdout.toString().trim();
             if (semver.gte(ionicVersion, '3.0.0')) {
                 args.unshift('cordova');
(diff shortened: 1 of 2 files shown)


src/debugger/extension.ts
```diff
@@ -79,11 +82,16 @@ export function cordovaStartCommand(args: string[], cordovaRootPath: string): ch
     let commandExtension = os.platform() === 'win32' ? '.cmd' : '';
     let command = cliName + commandExtension;
 
-    let isIonicServe:boolean = args.indexOf("serve") >= 0;
```
vladimir-kotikov: Don't we need to check whether it's a 'serve' command anymore? I mean in the past ionic serve didn't require prepending by "cordova" (see #283)

src/debugger/extension.ts
```diff
@@ -79,11 +82,16 @@ export function cordovaStartCommand(args: string[], cordovaRootPath: string): ch
...
     let command = cliName + commandExtension;
 
-    let isIonicServe:boolean = args.indexOf("serve") >= 0;
-
-    if (CordovaProjectHelper.isIonicCordovaCLINamespacedProject(cordovaRootPath) && !isIonicServe) {
-        // add cordova namespace to Ionic projects that uses cli-plugin-cordova
-        args.unshift('cordova');
+    if (cliName === 'ionic') {
+        try {
+            let ionicInfo = child_process.spawnSync(command, ['info']);
```
vladimir-kotikov: Instead of `ionic info` I'd rather use `ionic -v` command which will print only CLI version without need to use regular expressions. IMO it's format is also more stable and less likely to change in the future.

PR #310, opened 2017-08-21
Fix attach failure when both device and emulator are connected

### LINKED ISSUE #307: Extension fails to "find localabstract name" when there are both device and emulator are connected
Repro steps:
- Connect device to dev machine
- Run android emulator (100% reproduces with emulator from Android SDK)
  _Notice - order is important - you need to connect device first then run emulator_
- Open Cordova/Ionic project and start debugging using "Run Android on device" configuration

App builds and runs on device, but debugger fails to attach with the following message:

```
Package name: io.ionic.starter
LAUNCH SUCCESS
App successfully launched
Attaching to android
Unable to find localabstract name of cordova app
```

Tested on current master branch with #306 applied, so this seems to be another issue. From what  I can see the problem is that even app runs on device, debug adapted handles this incorrectly and tries to find appropriate debug port to connect to on simulator rather than on device.

vladimir-kotikov:
Determine device/emulator name to attach to using algorithm used by cordova-android. Also handle correctly values for "target" other than device/emulator.

This fixes #307

### DIFF
--- a/src/debugger/cordovaDebugAdapter.ts
+++ b/src/debugger/cordovaDebugAdapter.ts
@@ -304,40 +304,48 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
     }
 
     private attachAndroid(attachArgs: ICordovaAttachRequestArgs): Q.Promise<IAttachRequestArgs> {
-        let errorLogger = (message) => this.outputLogger(message, true);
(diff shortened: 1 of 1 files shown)


vladimir-kotikov:
@iTOYS can you take a look please?

iTOYS:
@vladimir-kotikov LGTM

PR #316, opened 2017-09-08
Correct splitting adb shell columns

iTOYS:
Fix for #312


vladimir-kotikov:
Tested, confirm it works fine w/ Oreo

PR #320, opened 2017-09-18
Use pidof command if applicable

iTOYS:
Fix for #312

### DIFF
--- a/src/debugger/cordovaDebugAdapter.ts
+++ b/src/debugger/cordovaDebugAdapter.ts
@@ -86,6 +86,8 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
     private static CHROME_DATA_DIR = 'chrome_sandbox_dir'; // The directory to use for the sandboxed Chrome instance that gets launched to debug the app
     private static NO_LIVERELOAD_WARNING = 'Warning: Ionic live reload is currently only supported for Ionic 1 projects. Continuing deployment without Ionic live reload...';
     private static SIMULATE_TARGETS: string[] = ['chrome', 'chromium', 'edge', 'firefox', 'ie', 'opera', 'safari'];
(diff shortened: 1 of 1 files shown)


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -346,67 +346,75 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
...
-                                return pathField.substr(1);
-                            }
 
-                            if (pathField === `@${appPackageName}_devtools_remote`) {
-                                // Matches the crosswalk format of "@PACKAGENAME_devtools_remote
-                                return pathField.substr(1);
-                            }
-                            // No match, keep searching
-                        }
+                        return this.runAdbCommand(getPidCommandArguments, errorLogger)
```
vladimir-kotikov: @iTOYS I think that in case if `pidof` doesn't exist on device `this.runAdbCommand(pidofCommandArguments, errorLogger)` will throw, so that you need to catch this error and fall back to old behavior - it's not sufficient just to check if `pidof` returned any result

iTOYS: `pidof` won't throw any errors, only this message if fault:
`/system/bin/sh: pidof: not found`
And [this check](https://github.com/Microsoft/vscode-cordova/pull/320/files#diff-b0f7f8daa1da800e681fd92bc82fd49eR357) to be sure that `pidof` returned PID

vladimir-kotikov: In case when `pidof` not found `adb shell` will exit with nonzero code which in turn would cause a promise rejection returned here https://github.com/Microsoft/vscode-cordova/blob/master/src/debugger/extension.ts#L30-L33

Please double check

iTOYS: `adb shell pidof` will exit with nonzero code only if `shell` command not found

PR #322, opened 2017-09-22
Run application on ios device with ios-deploy

iTOYS:


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -57,6 +57,7 @@ export interface ICordovaLaunchRequestArgs extends DebugProtocol.LaunchRequestAr
 export interface ICordovaAttachRequestArgs extends DebugProtocol.AttachRequestArguments, IAttachRequestArgs {
     cwd: string; /* Automatically set by VS Code to the currently opened folder */
     platform: string;
+    request?: string,
```
vladimir-kotikov: @iTOYS, looks like this property is unused?

iTOYS: You're right, weird...

vladimir-kotikov:
@iTOYS LGTM, but I'd like you to incorporate https://github.com/Microsoft/vscode-cordova/pull/325 into this branch as well

PR #324, opened 2017-10-06
Non interactive mode

iTOYS:
#323


vladimir-kotikov:
@iTOYS LGTM. The only my concern is whether this will work with previous Ionic/Cordova versions. Which Cordova and Ionic versions did you tested this with?

iTOYS:
@vladimir-kotikov Works fine with cordova@5/ionic@2 or greater

vladimir-kotikov:
Awesome!

PR #326, opened 2017-10-13
Fix missing platform error message for Ionic

janpio:
Ionic CLI command is `ionic cordova platform add ...`

### DIFF
--- a/src/extension/simulate.ts
+++ b/src/extension/simulate.ts
@@ -75,7 +75,7 @@ export class PluginSimulator implements vscode.Disposable {
                 if (isPlatformMissing) {
                      let command = "cordova";
                      if (projectType.ionic || projectType.ionic2) {
-                        command = "ionic";
+                        command = "ionic cordova";
                      }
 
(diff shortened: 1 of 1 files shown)


iTOYS:
Hi @janpio,
Thanks for contributing! But it is is not work with Ionic@2

janpio:
If you refer to Ionic CLI, yes, this does not work with Ionic CLI 2.x. But the current version has been 3.x now for quite some time. The current output does not work with Ionic CLI 3.x.

(Does the rest of this extension use `ionic cordova` commands for everything?)

iTOYS:
Extension use `ionic cordova` command for Ionic CLI@3 or greater otherwise just `ionic` command.
https://github.com/Microsoft/vscode-cordova/blob/master/src/debugger/extension.ts#L110L112

janpio:
Ahh. Ok, so the same should be done to this error message then, correct?

iTOYS:
Yes, something like that.

svekl:
@iTOYS Please, close this PR, fixed this message in #366

Issue #281, opened 2017-06-12
F6 Build config


andreujuanc:
Hello,

I could not find docs to configure the behaviour of Cordova Build. Is this possible?

It'd be nice to have, like we have launch.json, a build.json, where you can set defaults, presets etc.

For example, now i'm developing a hook that works on windows and android. I have to bring the console up, arrow up, enter. All the time. I'd lose less time and be more productive if I could just press F6 and set a default build platform and architecture. This is because i'm using certain plugins that don't allow me to build on AnyCPU.

Is this possible? Should I just modify cordova.cmd?

vladimir-kotikov:
See also #250.

As a workaround you can use a concept of [vscode tasks](https://code.visualstudio.com/docs/editor/tasks) to run any arbitrary command (including `cordova build`) and bind it to hotkey of your choice.

As a side note - IMHO tasks are really powerful and even more appropriate for this than having a special command, that just builds app

iTOYS:
Implemented in #328

Issue #219, opened 2016-12-21
Support "skipFiles" options in launch.json to avoid debugging certain files/directories


gianpaj:
Also called "just my code"

https://code.visualstudio.com/Docs/editor/debugging#_just-my-code-skipfiles

MSLaguana:
This should be supported once we upgrade to the newest vscode-chrome-debug-core I believe.

iTOYS:
Implemented in #328
Added 'runArguments' options

hqp6684:
@iTOYS if is it possible, can you please give me an example for skipping files ? 
This is what I have but it doesn't seem to work
`{
  "configurations": [
    {
      "name": "Run android|ios on device|emulator",
      "type": "cordova",
      "request": "launch",
      "platform": "android",
      "target": "emulator",
      "sourceMaps": true,
      "cwd": "${workspaceRoot}",
      "livereload": true,
      // "ionicLiveReload": true
      "runArguments": ["skipFiles", "core.es5.js"]
    }
  ]
}
`

svekl:
Hi @hqp6684,
looks like some misunderstanding occurred, you can pass any parameters to chrome debugger, including `skipFiles`, you might be confused of lack of autocomplete for this parameter, we will add it in future version, but it will be passed to chrome debugger anyway. So your configuration might look like that
```
    {
      "name": "Run android|ios on device|emulator",
      "type": "cordova",
      "request": "launch",
      "platform": "android",
      "target": "emulator",
      "sourceMaps": true,
      "cwd": "${workspaceRoot}",
      "livereload": true,
      // "ionicLiveReload": true
      "skipFiles": ["core.es5.js"]
    }
```

not sure if you need to add path to `core.es5.js` or not, depends on your project configuration, but chrome debugger uses glob pattern for `skipFiles` so you may use it if you need.

PR #328, opened 2017-10-18
Implement runArguments in settings.json

svekl:
Adds `runArguments` configuration property to set arguments to be passed to `cordova/ionic run/build` or `ionic serve` command


iTOYS:
#281 #219

PR #335, opened 2017-11-07
Add skipFiles debugger parameter

svekl:


svekl:
@iTOYS Please, take a look

PR #345, opened 2017-11-19
Added runArguments to documentation

markacianfrani:
I noticed runArguments (super helpful btw) wasn't documented anywhere so I just made a quick note.


iTOYS:
Thanks Mark!

PR #350, opened 2017-11-27
Fix ionic nightly version detection

svekl:
As reported in #134, Ionic version detection fails for `nightly`. This PR assumes that `nightly` and `latest` is Ionic > 3.0.0


Kaidanov:
Where can I find this files on my pc and change them? 
A bit lost..

Kaidanov:
Found it at C:\Users\{MYUSER}\.vscode\extensions\vsmobile.cordova-tools-1.3.1\out\src\utils
Works like a charm. Thanks a lot!

PR #358, opened 2018-01-09
Support new path of AndroidManifest.xml in Cordova 8.0.0

### LINKED ISSUE #355: AndroidManifest.xml not found using Cordova-Android 7.0.0
# My setup:

**Vscode:** 1.19.1
**Cordova Tools:** 1.3.1
**Cordova version:** 8.0.0
**Cordova-Android version:** 7.0.0
**Launch.json**: https://gist.github.com/SpellCraft/271e4f8e121226a091852e64d25d7317

# Issue:
Running Cordova on android emulator with the latest version of everything works fine, until the debugger is attempting to attach itself to the newly launched instance. It's generating this error:

ENOENT: no such file or directory, open '...\platforms\android\AndroidManifest.xml'

As far as I'm able to deduct is that Cordova changed the folder structure on Android on the 7.0.0 launch such so that AndroidManifest.xml is no longer found in the base /Cordova Project/Platforms/Android/ folder, 

I think it was moved to platforms/android/app/src/main as that's the only place other than outside the CordovaLib folder I can find it.

iTOYS:
Fixes #355

### DIFF
--- a/src/debugger/cordovaDebugAdapter.ts
+++ b/src/debugger/cordovaDebugAdapter.ts
@@ -28,6 +28,8 @@ import {Telemetry} from '../utils/telemetry';
 import {SimulationInfo} from '../common/simulationInfo';
 
 const MISSING_API_ERROR = 'Debugger.setAsyncCallStackDepth';
+const ANDROID_MANIFEST_PATH = path.join('platforms', 'android', 'AndroidManifest.xml');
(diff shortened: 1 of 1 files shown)


iTOYS:
@svekl PTAL

svekl:
@iTOYS It might be better to check if file exists on both paths, and open found one, instead of catching error

iTOYS:
@svekl I guess that check if file exists are redundant. Also this is work a bit faster

Issue #87, opened 2016-03-31
Debug fails with [cordova-debug-adapter] Error: Unable to find webview


indy7:
I just updated my VSC with the latest 1.1.0 cordova-tools.
When I debug my Ionic2 project w/ Run on iOS Simulator, I keep getting the '[cordova-debug-adapter] Error: Unable to find webview'.
I am running mac OS El Capitan and XCode 7.3.
From my observation, this error occurs while it is trying to load the app in the simulator.
The thing is, even with this error the app eventually loads in the simulator.
But in VSC the debug has ended due to the error. So I have no way of getting the breakpoint to hit, etc.

MSLaguana:
How long did your app take to launch? If it takes longer than 5 seconds to start up you may need to increase the timeout in your launch.json by specifying `attachAttempts` or `attachDelay`. By default we try 5 times with a delay of 1000ms between each attempt.

imacnu:
Same erro "Unable to find webview", can you an example aatachDelay or/and atatchAttemps configuration?

imacnu:
with configuration:

{
			"name": "Attach to running iOS on simulator",
			"type": "cordova",
			"request": "attach",
			"platform": "ios",
			"target": "emulator",
			"port": 9222,
			"sourceMaps": true,
			"cwd": "${workspaceRoot}",
			"attachAttempts": 5,
			"attachDelay": 5000
		},


Same error... Unable to find webview

SimpleDeveloper49:
Just as an FYI, I used to run into the same issue and using attachDelay and attachAttempts did resolve this for me. 

@imacnu I used the snippet you provided and it works as expected for me.

guillaumejenkins:
@imacnu This error can also happen if you have more than one Simulator running (for example, if several users are logged in on the same machine and each using the Simulator). If that's the case for you, you will need to fully quit all the other simulators.

If that's not it, then how long does your app take to fully launch on the simulator? Your example shows 5 attempts of 5 seconds each; maybe that's not enough? Is your app fully up and running within 25 seconds? If not, it may be worth increasing the attachDelay to a higher value to see if that helps.

Let us know how that goes, and if it didn't help we'll investigate further.

imacnu:
Still does not work with any delay :(

MSLaguana:
[...]

PR #359, opened 2018-01-10
Fix Ionic 1 livereload server detection

svekl:
Ionic 1 has slightly different output then we expect

Fixes issue mentioned in
https://github.com/Microsoft/vscode-cordova/issues/87#issuecomment-355715039


svekl:
@iTOYS Please, take a look

PR #365, opened 2018-01-23
Fixes error message if browser simulation fails

svekl:
`simulator.startSimulation()` doesn't reject if simulation start fails


svekl:
@iTOYS Please, take a look

PR #366, opened 2018-01-23
Fix missing platform error message for Ionic

svekl:


svekl:
@iTOYS Please, take a look

PR #377, opened 2018-03-01
Cleaning up UI strings.

pniko:


src/extension/appcenter/appCenterStrings.ts
```diff
@@ -26,7 +26,7 @@ export class ACStrings {
     public static FetchDeploymentsStatusBarMessage: string = 'Fetching app deployments for you...';
     public static GettingAppInfoMessage: string = 'Getting app info...';
     public static DetectingAppVersionMessage: string = 'Detecting app version...';
-    public static RunningReactNativeBundleCommandMessage: string = 'Running \"react-native bundle\" command...';
+    public static RunningBundleCommandMessage: string = 'Running bundle command...';
```
pniko: @max-mironov please review the changes.

max-mironov: @pniko need to replace this string https://github.com/Microsoft/vscode-cordova/blob/master/src/extension/appcenter/command/commandExecutor.ts#L245 then, no probs, I'll take care of it in separate commit!

max-mironov:
@pniko Im merging this for Cordova extension, do we need to update the corresponding strings for ReactNative extension?

sergey-akhalkov:
@max-mironov, my understanding is that `appcenter` folder contains almost the same code for both extensions, so I believe that this folder should be as generic as possible to make it simpler to synchronize and maintain it in the future.

max-mironov:
Ya, in this case not generic but probably almost the same (except cordova/react specific parts), I'll take care of it!

PR #379, opened 2018-03-05
Fix dependencies

ruslan-bikkinin:
This PR fixes dependencies errors in extension during runtime.


ruslan-bikkinin:
@sergey-akhalkov agreed, it should be done in the next release.

PR #399, opened 2018-05-10
Delete unused code

iTOYS:
cc/ @ruslan-bikkinin

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -52,7 +52,7 @@
     "no-keyword-named-variables": true,
     "no-require-imports": false,
     "no-shadowed-variable": true,
-    "no-string-literal": true,
+    "no-string-literal": false,
     "no-switch-case-fall-through": true,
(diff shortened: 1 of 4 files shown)


tslint.json
```diff
@@ -52,7 +52,7 @@
     "no-keyword-named-variables": true,
     "no-require-imports": false,
     "no-shadowed-variable": true,
-    "no-string-literal": true,
+    "no-string-literal": false,
```
ruslan-bikkinin: Are you sure we should get rid of it?

iTOYS: Yes, we use access by string literal in telemetry props

https://github.com/Microsoft/vscode-cordova/blob/003b629914153e5752ba8e34063a551f013db72b/src/utils/telemetryHelper.ts#L16-L18

ruslan-bikkinin: @iTOYS okay then

PR #405, opened 2018-05-15
Strict detect target type for telemetry

iTOYS:


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -1278,6 +1278,21 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
         }
     }
 
+    /**
+     * Target type for telemetry
+     */
+    private static getTargetType(target: string): string {
+        if (/emulator/i.test(target)) {
+            return 'emulator';
```
ruslan-bikkinin: I believe we should add `TargetType` enum for this instead of using inline strings.

iTOYS: Done

PR #407, opened 2018-05-25
Selection target platform for command palette commands

### LINKED ISSUE #250: 'Ionic: Build' and 'Ionic: Run' commands should allow user to specify platform

iTOYS:
Resolve #250

### DIFF
--- a/src/utils/cordovaCommandHelper.ts
+++ b/src/utils/cordovaCommandHelper.ts
@@ -99,27 +99,7 @@ export class CordovaCommandHelper {
 
     private static selectPlatform(projectRoot, command): Q.Promise<string> {
         let platforms = CordovaProjectHelper.getInstalledPlatforms(projectRoot);
-
-        if (os.platform() === 'win32') {
-            platforms = platforms.filter((platform) => {
-                switch (platform) {
-                    case 'ios':
-                    case 'osx':
-                        return false;
-                    default:
-                        return true;
-                }
-            });
-        } else {
-            platforms = platforms.filter((platform) => {
-                switch (platform) {
-                    case 'windows':
-                        return false;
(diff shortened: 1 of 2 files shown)


src/utils/cordovaCommandHelper.ts
```diff
@@ -95,4 +96,56 @@ export class CordovaCommandHelper {
...
+                    if (platforms.length > 1) {
+                        platforms.unshift('all');
+                        return window.showQuickPick(platforms)
+                            .then((platform) => {
+                                if (!platform) {
+                                    throw new Error('Platform selection was canceled. Please select target platform to continue!');
+                                }
+
+                                if (platform === 'all') {
+                                    return '';
```
ruslan-bikkinin: I believe it could be replaced with `return null;`due to [this change](https://github.com/Microsoft/vscode-cordova/pull/407/files?utf8=%E2%9C%93&diff=split&w=1#diff-f060c4215ef32ba898c255bbdd0eea0cR40).

iTOYS: `'' == false`

PR #421, opened 2018-08-02
Added support for cordovaExecutable, env, envFile as a launch params, settings

### LINKED ISSUE #181: Add support for local ionic/cordova installs
Hi, please add support for using local per-project installs of ionic/cordova, if they exist. Currently, the extension expects a single version of each to be installed globally and on the system path, which makes it difficult to use across projects spanning multiple versions of each. Thanks!

### LINKED ISSUE #415: Add support for environment variables
If this is already implemented, please let me know.

Currently, I can run an ionic project with the following commands to change environment:

`ENV=staging ionic serve -l`
`ENV=dev ionic cordova run android -l`
`ENV=production ionic cordova build android`

I want to be able to do this using the debugger, but I can't figure out how to pass that variable in when starting the debugger. Any help?

iTOYS:
Resolves #181 
Resolves #415

### DIFF
--- a/package.json
+++ b/package.json
@@ -399,24 +399,24 @@
                 "description": "Path to local cordova/ionic executable"
               },
               "env": {
-								"type": "object",
-								"additionalProperties": {
-									"type": "string"
-								},
-								"description": "Environment variables passed to the program.",
-								"default": {}
-							},
-							"envFile": {
-								"type": "string",
-								"description": "Absolute path to a file containing environment variable definitions.",
-								"default": "${workspaceFolder}/.env"
-							},
+                "type": "object",
+                "additionalProperties": {
+                  "type": "string"
+                },
+                "description": "Environment variables passed to the program.",
+                "default": {}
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -394,6 +394,23 @@
                 "description": "Launch Cordova simulator in browser instead of tab in VS Code",
                 "default": false
               },
+              "cordovaExecutable": {
+                "type": "string",
+                "description": "Path to local cordova/ionic executable"
+              },
+              "env": {
+								"type": "object",
```
ruslan-bikkinin: Please check indents - looks like there are too many of them.

maefinho:
Hey Guys,

sorry for the stupid question but where do I have to set my env variables now exactly? In the launch.json configruations sections or where?

I found nothing in the docs. I am using cordova-tools 1.5.1 and this feature was part of the 1.5.0 release. 

Thanks for your help and best regards

Issue #426, opened 2018-08-19
Ionic v4 breakpoints are unverified


dinana:
## Actual Behavior
1. Select 'Simulate iOS in browser' in VS Code
2. Set breakpoint
3. breakpoint is unverified

## Expected Behavior
1. breakpoint is verified
2. breakpoint is hit

## Software versions
- Cordova VS Code extension version: 1.4.0
- VSCode version: 1.26.1
- OS platform and version: MacOS 10.13.2
- NodeJS version: 10.6.0
- Cordova version: 8.0.0
- Cordova-Android/iOS/Windows version: 

## Outputs (Include if relevant)
- Output of the Debug Console (View -> Toggle Debug Console):
``` 
Error processing "setBreakpoints": TypeError: Cannot read property 'mappedPath' of undefined
```
- Output of the Cordova output channel (View -> Toggle Output -> Select Cordova in ListBox):
```
-
```

dinana:
Worth mentioning, that if I enable developer tools in chrome and set a breakpoint there it is hit and VS Code is focused with the breakpoint hit, this is how it shows in the breakpoints panel:
<img width="529" alt="image" src="https://user-images.githubusercontent.com/1944157/44310690-c948f580-a3da-11e8-8f03-999c4311daaa.png">

U can see the old breakpoint set in my source shown as unverified and the one from chrome beneath it.

bidwej:
How did you hit breakpoints in Chrome? Did you set typescript mapping paths in "tsconfig.json"? Thank you!

bidwej:
Thank you! The suggested solution worked for me.

PR #432, opened 2018-10-11
Fixing bugs with ionic 4 breakpoints shows unverified

SounD120:
Fix for #426
Updated Chrome debug core version

### DIFF
--- a/package.json
+++ b/package.json
@@ -572,7 +572,7 @@
     "source-map-support": "^0.4.18",
     "temp": "^0.8.3",
     "vscode": "^1.1.6",
-    "vscode-chrome-debug-core": "^6.7.23",
+    "vscode-chrome-debug-core": "^6.7.25",
     "vscode-debugadapter": "^1.32.1",
     "vscode-extension-telemetry": "0.0.5",
     "which": "^1.3.0",
--- a/package.json
+++ b/package.json
@@ -108,16 +108,7 @@
             "port": 9222,
             "sourceMaps": true,
             "cwd": "${workspaceFolder}",
-            "ionicLiveReload": false,
-            "sourceMapPathOverrides": {
-              "webpack:///./~/*": "${cwd}/node_modules/*",
-              "webpack:///./*": "${cwd}/*",
-              "webpack:///*": "*",
-              "webpack:///src/*": "${cwd}/*"
-            },
-            "pathMapping": {
-              "/": "${cwd}/*"
-            }
+            "ionicLiveReload": false
           },
           {
             "name": "Run iOS on device",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -538,8 +680,8 @@
     "source-map-support": "^0.4.18",
     "temp": "^0.8.3",
     "vscode": "^1.1.6",
-    "vscode-chrome-debug-core": "3.21.2",
-    "vscode-debugadapter": "^1.22.0",
+    "vscode-chrome-debug-core": "^6.7.19-patch.0",
```
dhanvikapila: Should we pick up the latest version `6.7.23`

package.json
```diff
@@ -107,8 +107,17 @@
             "target": "device",
             "port": 9222,
             "sourceMaps": true,
-            "cwd": "${workspaceRoot}",
-            "ionicLiveReload": false
+            "cwd": "${workspaceFolder}",
+            "ionicLiveReload": false,
+            "sourceMapPathOverrides": {
```
ruslan-bikkinin: Can we override sourcemap paths implicitly inside debugging session for ionic applications without forcing users to change their configs?

src/debugger/cordovaPathTransformer.ts
```diff
@@ -34,98 +25,98 @@ export class CordovaPathTransformer extends BasePathTransformer {
...
-                resolve();
-                return;
-            }
+    public setBreakpoints(args: ISetBreakpointsArgs): ISetBreakpointsArgs {
+        if (!args.source.path) {
+            // sourceReference script, nothing to do
+            return args;
+        }
 
-            const url = utils.canonicalizeUrl(args.source.path);
```
ruslan-bikkinin: Why this block is removed? Is it now inside BasePathTransformer?

src/debugger/cordovaIosDeviceLauncher.ts
```diff
@@ -229,7 +229,7 @@ export class CordovaIosDeviceLauncher {
     private static mountDeveloperImage(): Q.Promise<any> {
         return CordovaIosDeviceLauncher.getDiskImage()
             .then(function (path: string): Q.Promise<any> {
-            let imagemounter: child_process.ChildProcess = child_process.spawn("ideviceimagemounter", [path]);
+            let imagemounter: child_process.ChildProcess = child_process.spawn("ideviceimagemounter", [path, path + ".signature"]);
```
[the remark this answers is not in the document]
SounD120: I looked [this docs](http://docs.quamotion.mobi/en/latest/imobiledevice/ideviceimagemounter.html) and found out that the signature file stores like this
image
so i added this argument

src/debugger/cordovaPathTransformer.ts
SounD120: I replaced all code from old path transformer by [Url path transformer](https://github.com/Microsoft/vscode-chrome-debug-core/blob/master/src/transformers/urlPathTransformer.ts) and adapted it to cordova tools

PR #462, opened 2018-12-04
Fix security vulnerabilities (except for cordova-simulate)

ruslan-bikkinin:
This PR fixes all security vulnerabilities except the ones caused by cordova-simulate and its sub dependencies.


gulpfile.js
```diff
@@ -62,6 +62,8 @@ var tsConfig = require('./tsconfig.json');
 var projectConfig = tsConfig.compilerOptions;
 projectConfig.typescript = typescript;
 
+let watching = false;
```
SounD120: Do we need this variable?

ruslan-bikkinin: Thanks for pointing, already fixed.

PR #474, opened 2019-01-11
Replaced unlink by unlinkSync to avoid potential problems with VS Code v1.31

SounD120:
fixes situation described at gruntjs/grunt-contrib-jasmine#266
AB#307


src/utils/tsdHelper.ts
```diff
@@ -79,7 +79,7 @@ export class TsdHelper {
         if (typeDefsToRemove.length === 0) return;
 
         typeDefsToRemove.forEach(typeDef => {
-            fs.unlink(path.resolve(typingsFolderPath, typeDef), err => {
+            fs.unlinkSync(path.resolve(typingsFolderPath, typeDef), err => {
```
ruslan-bikkinin: Seems to that this unlink is asynchronous, so no need to change it.

PR #497, opened 2019-02-05
Prepare for 1.8.0

ruslan-bikkinin:


LiamKarlMitchell:
Will 1.8.0 include the fix for renamed commands?

```The run command has been renamed. To find out more, run:

     ionic cordova run --help```

ruslan-bikkinin:
@LiamKarlMitchell thanks for pointing to that! Yeah, we'll include fix for it, as well. Is it for `Ionic: Run`  palette command only, I assume?

CodeDoctor34:
Any idea when the next release will be available?

ruslan-bikkinin:
@CodeDoctor34 1.8.0 is just been released.

PR #507, opened 2019-02-13
Fix debugging with livereload option is enabled and www dir is not empty

ruslan-bikkinin:
AB#423


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -40,6 +40,9 @@ export interface ICordovaAttachRequestArgs extends DebugProtocol.AttachRequestAr
     attachAttempts?: number;
     attachDelay?: number;
     simulatorInExternalBrowser?: boolean;
+
+    // Ionic livereload properties
```
```suggestion
    // Ionic livereload property
```

SounD120: Actually, do we need that comment?

PR #509, opened 2019-02-13
Fix attaching to ionic 4 apps with livereload

ruslan-bikkinin:
AB#423

### DIFF
--- a/package.json
+++ b/package.json
@@ -351,7 +351,7 @@
               },
               "ionicLiveReload": {
                 "type": "boolean",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -496,6 +496,11 @@
                   "/": "${workspaceFolder}"
                 }
               },
+              "ionicLiveReload": {
+                "type": "boolean",
+                "description": "Set this to true if you are going to attach to app being ran with --livereload option",
```
SounD120:
```suggestion
                "description": "Set this to true if you are going to attach to Ionic app being ran with --livereload option",
```

PR #518, opened 2019-04-09
Bump core dependencies

SounD120:
AB#571

### DIFF
--- a/src/debugger/cordovaPathTransformer.ts
+++ b/src/debugger/cordovaPathTransformer.ts
@@ -70,8 +70,8 @@ export class CordovaPathTransformer extends BasePathTransformer {
         this._targetUrlToClientPath = new Map<string, string>();
     }
 
-    public scriptParsed(scriptUrl: string): Promise<string> {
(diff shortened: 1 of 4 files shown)


src/debugger/cordovaPathTransformer.ts
```diff
@@ -145,19 +145,17 @@ export class CordovaPathTransformer extends BasePathTransformer {
...
-                "/": `${searchFolder}`,
+                "/": searchFolder,
             };
-            let mappedPath = chromeUtils.targetUrlToClientPath(sourceUrl, pathMapping);
+            let mappedPath = await chromeUtils.targetUrlToClientPath(sourceUrl, pathMapping);
 
             if (mappedPath) {
                 defaultPath = mappedPath;
-                return true;
+                break;
```
ruslan-bikkinin: Could you please clarify why this change is needed?

SounD120: There is an asynchrounous await in predicate function. It was easier to refactor this function instead of returning lots of promises

PR #522, opened 2019-04-11
Fix iOS debugging for Cordova 9 app

### LINKED ISSUE #521: Error "Application not installed on the device" while debugging Cordova 9 iOS apps
Error "Application not installed on the device" occurred while debugging Cordova 9 iOS apps. The cause of issue is change in cordova-ios@5.0.0: instead of using plain CFBundleIdentifier value in .plist, `$(PRODUCT_BUNDLE_IDENTIFIER)` is used now.

ruslan-bikkinin:
Fixes #521


package.json
```diff
@@ -1,592 +1,598 @@
...
-    "mockery": "^1.4.0",
-    "rimraf": "^2.6.3",
-    "should": "^13.2.1",
-    "tslint": "^5.6.0",
-    "typescript": "2.4.2",
-    "vsce": "^1.3.0",
-    "vscode": "^1.1.29"
-  }
-}
+	"name": "cordova-tools",
```
SounD120: Please revert this file

SounD120: Diff is still looks dirty, I think it would be better to revert the whole file and then add packages you need. It'll make clean and understandable diff.

package.json
```diff
@@ -589,4 +590,4 @@
     "vsce": "^1.3.0",
     "vscode": "^1.1.29"
   }
-}
+}
```
SounD120:
```suggestion
}

```

ruslan-bikkinin:
```suggestion
}
```

test/debugger/cordovaIosDeviceLauncher.ts
```diff
@@ -13,7 +13,7 @@ describe("cordovaIosDeviceLauncher", function () {
     let fsMock: any = {};
 
     before(() => {
-        mockery.enable({warnOnReplace: false, useCleanCache: true});
+        mockery.enable({warnOnReplace: false, useCleanCache: true, warnOnUnregistered: false});
```
ruslan-bikkinin:
```suggestion
        // warnOnUnregistered is set to false because of https://github.com/mfncooper/mockery/issues/59
        mockery.enable({warnOnReplace: false, useCleanCache: true, warnOnUnregistered: false});
```

PR #583, opened 2019-10-14
Fix inline breakpoints behavior on iOS debugging

RedMickey:
1. Disable inline breakpoints for iOS app debugging on device
1. Rename `setBreakpoints` function input parameter from `args` to `source`


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -79,6 +79,12 @@ export interface ICordovaCommonRequestArgs extends ICommonRequestArgs {
     cwd?: string;
 }
 
+interface DebuggingProperties {
+    // Workaround to handle breakpoint location requests correctly on some platforms
```
```suggestion
```

SounD120: I think one mention below will be enough.

PR #600, opened 2019-12-26
Fix Run iOS on device scenario for iOS 13+

RedMickey:


src/debugger/cordovaDebugAdapter.ts
```diff
@@ -650,89 +650,39 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
...
-            if (projectType.ionic || projectType.ionic2 || projectType.ionic4)
-                args = ["build", "ios", "--", "--buildFlag=-UseModernBuildSystem=0"];
+            let args = ["run", "ios", "--device", "--buildFlag=-UseModernBuildSystem=0"];
 
             if (launchArgs.runArguments && launchArgs.runArguments.length > 0) {
                 args.push(...launchArgs.runArguments);
             } else if (runArguments && runArguments.length) {
                 args.push(...runArguments);
-            } else {
-                args.push("--device");
```
SounD120: Could you please explain why we don't need this argument here? Is this expected change?

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -650,89 +650,39 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
...
             }
 
             if (args.indexOf("--livereload") > -1) {
-                if (args[0] === "build") {
-                    args[0] = "run";
-                }
                 return this.startIonicDevServer(launchArgs, args).then(() => void 0);
             }
 
             // cordova run ios does not terminate, so we do not know when to try and attach.
```
SounD120: I think it would be nice to expand this comment a bit, e.g. add info about our output parsing approach

src/debugger/cordovaDebugAdapter.ts
RedMickey: This argument has been removed since the same one already exists [here](https://github.com/microsoft/vscode-cordova/pull/600/files#diff-b0f7f8daa1da800e681fd92bc82fd49eR657).

src/debugger/cordovaDebugAdapter.ts
RedMickey: Additional description was added.

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -650,89 +650,40 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
...
-            }, undefined, (progress) => {
-                this.outputLogger(progress[0], progress[1]);
-            }).spread((appBundleId) => {
-                // App is now built and installed. Try to launch
-                this.outputLogger("Launching app");
-                return CordovaIosDeviceLauncher.startDebugProxy(iosDebugProxyPort).then(() => {
-                    return CordovaIosDeviceLauncher.startApp(appBundleId, iosDebugProxyPort, appStepLaunchTimeout);
-                });
-            }).then(() => void (0));
+            // Therefore we parses the command's output to find the special key, which means that the application has been successfully launched.
```
```suggestion
            // Therefore we parse the command's output to find the special key, which means that the application has been successfully launched.
```

src/debugger/cordovaDebugAdapter.ts
```diff
@@ -650,89 +650,40 @@ export class CordovaDebugAdapter extends ChromeDebugAdapter {
         this.outputLogger("Launching app (This may take a while)...");
 
         let iosDebugProxyPort = launchArgs.iosDebugProxyPort || 9221;
-        let appStepLaunchTimeout = launchArgs.appStepLaunchTimeout || 5000;
 
         const command = launchArgs.cordovaExecutable || CordovaProjectHelper.getCliCommand(workingDirectory);
         // Launch the app
         if (launchArgs.target.toLowerCase() === "device") {
-            // Workaround for dealing with new build system in XCode 10
-            // https://github.com/apache/cordova-ios/issues/407
```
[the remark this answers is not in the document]
RedMickey: It seems that this is relevant comment, I'll restore it.

Issue #613, opened 2020-04-20
Migrate on js-debug extension's debugging core


SounD120:
We are planning to migrate Cordova extension debugging core to [js-debug](https://github.com/microsoft/vscode-js-debug) extension one, since currently used [vscode-chrome-debug-core](https://github.com/microsoft/vscode-chrome-debug-core) tool will be deprecated soon. For providing communication between app and `js-debug` extension the special [CDP proxy](https://github.com/microsoft/vscode-cdp-proxy) will be used. It will allow to adapt platform specific features on the debugging protocol level.

Below there is the migration roadmap for tracking the progress.

# Roadmap

- [x] **[proxy]** Implement translation of a target URL to Websocket inspector URL
- [x] **[extension]** Implement Debug Configuration Provider
- [x] **[extension]** Implement running of the debug adapter directly inside the extension and make VS Code to connect to it instead of launching a new external debug adapter per session
- [x] **[extension]** Implement launch of CDP proxy
- [x] **[extension]** Implement starting of js-debug extension's debug session
- [x] **[extension]** Implement multiroot workspaces support
- [x] **[proxy]** Add listeners on CDP events inside the proxy
- [x] **[proxy]** Implement correct path resolving 
- [x] **[extension]** Adapt current implementation of pre-attach setup for Android emulators and devices to proxy approach
- [x] **[extension]** Adapt current implementation of pre-attach setup for iOS simulator to proxy approach 
- [x] **[proxy]** Implement translation of flat session methods to non-flat ones if it's possible (to provide communication with IWDP)
- [x] **[ios-debug]** Investigate possible solutions for #564 (example https://github.com/RemoteDebug/remotedebug-ios-webkit-adapter/pull/177)
- [x] **[extension]** Adapt current implementation of Ionic live-reload handling to proxy approach
- [x] **[extension]** Сover changes with tests
- [x] **[extension]** Test the updated extension to make sure that all debug arguments work well
- [x] **[extension]** Migrate from TSLint to ESLint

SounD120:
Cordova Tools v2.0.0 with js-debug debugger have been released just now.

PR #614, opened 2020-04-23
Prepare for migration to js-debug

SounD120:
Refactor the extension code to make it use js-debug and vscode-cdp-proxy
Remove extension server, add DebugAdapterDescriptorFactory approach
#613


src/debugger/cordovaDebugSession.ts
```diff
@@ -30,67 +133,341 @@ export interface ILaunchRequestArgs extends DebugProtocol.LaunchRequestArguments
...
-        this.isSettingsInitialized = false;
-        this.appWorker = null;
-        this.rnCdpProxy = null;
-        this.debugSessionStatus = DebugSessionStatus.FirstConnection;
+        // this.isSettingsInitialized = false;
+        this.cordovaCdpProxy = null;
+        // this.debugSessionStatus = DebugSessionStatus.FirstConnection;
+        // this.cordovaPathTransformer = (<any>global).cordovaPathTransformer;
+        this.telemetryInitialized = false;
+        this.outputLogger = (message: string, error?: boolean | string) => {
```
RedMickey, line 188: Why do we need `outputLogger`? We can use `logger` from `vscode-debugadapter` package for logging.

src/debugger/cordovaDebugSession.ts
```diff
@@ -30,67 +133,341 @@ export interface ILaunchRequestArgs extends DebugProtocol.LaunchRequestArguments
...
+                        default:
+                            generator.add("unknownPlatform", platform, true);
+                            throw new Error(`Unknown Platform: ${platform}`);
+                    }
+                }).then((processedAttachArgs: IAttachRequestArgs & { url?: string }) => {
+                    this.outputLogger("Attaching to app.");
+                    this.outputLogger("", true); // Send blank message on stderr to include a divider between prelude and app starting
+                    return new Promise((resolve, reject) => {
+                        this.establishDebugSession();
+                        resolve();
```
RedMickey, line 430: It seems that we resolve attach promise without waiting of establishing connection to `js-debug`.

src/debugger/cordovaDebugSession.ts
```diff
@@ -115,30 +491,986 @@ export class CordovaDebugSession extends LoggingDebugSession {
...
-                    if (resolve) {
-                        this.debugSessionStatus = DebugSessionStatus.ConnectionAllowed;
-                        resolve();
-                    }
-                } else {
-                    this.debugSessionStatus = DebugSessionStatus.ConnectionFailed;
-                    throw new Error("Cannot start child debug session");
-                }
+            .then(() => {
+                // if (childDebugSessionStarted) {
```
RedMickey, line 500: I think we can resolve attach promise here.

src/debugger/cordovaDebugSession.ts
SounD120: Yes we can use it, but it will lead to huge changes in the rest of the code, so I think it is not the target of this PR.

src/debugger/cordovaDebugSession.ts
SounD120: Thanks, I'll take a look

PR #625, opened 2020-05-19
Fix multiroot projects processing

SounD120:
Example: 
image

[AB#955](https://dev.azure.com/vscode-webdiag-extensions/5c2c3798-7e49-49f6-b45f-77d9c69636c6/_workitems/edit/955)


src/debugger/cordovaDebugSession.ts
```diff
@@ -280,6 +226,9 @@ export class CordovaDebugSession extends LoggingDebugSession {
         // };
 
         return new Promise<void>((resolve, reject) => this.initializeTelemetry(launchArgs.cwd)
+            .then(() => {
+                this.workspaceManager = CordovaWorkspaceManager.getWorkspaceManagerByProjectRootPath(launchArgs.cwd);
```
RedMickey, line 230: I suppose, it could be better to move this call to initialisation part, since now, in case of launch scenario, we initialize it twice.

PR #636, opened 2020-07-14
Enhance js-debug debugging configuration

RedMickey:


src/debugger/cordovaDebugSession.ts
```diff
@@ -103,24 +111,6 @@ export interface IStringDictionary<T> {
     [name: string]: T;
 }
 export type ISourceMapPathOverrides = IStringDictionary<string>;
```
SounD120, line 113: I think it doesn't make sense to store this type and `IStringDictionary` interface in here. Lets put it in debug configuration adapter, or in the other file

src/debugger/cordovaDebugSession.ts
```diff
@@ -644,7 +628,7 @@ export class CordovaDebugSession extends LoggingDebugSession {
             // cordova run ios does not terminate, so we do not know when to try and attach.
             // Therefore we parse the command's output to find the special key, which means that the application has been successfully launched.
             this.outputLogger("Installing and launching app on device");
-            return cordovaRunCommand(command, args, launchArgs.env, workingDirectory)
+            return cordovaRunCommand(command, args, launchArgs.allEnv, workingDirectory)
```
SounD120, line 631: Why we rename that arg to allEnv?

src/debugger/cordovaDebugSession.ts
```diff
@@ -40,10 +41,21 @@ import { SourcemapPathTransformer } from "./cdp-proxy/sourcemapPathTransformer";
 const ANDROID_MANIFEST_PATH = path.join("platforms", "android", "AndroidManifest.xml");
 const ANDROID_MANIFEST_PATH_8 = path.join("platforms", "android", "app", "src", "main", "AndroidManifest.xml");
 
-export interface ICordovaAttachRequestArgs extends DebugProtocol.AttachRequestArguments, IAttachRequestArgs {
+export interface ICordovaAttachRequestArgs extends DebugProtocol.AttachRequestArguments {
```
SounD120, line 44: Lets put the interfaces like this one in other file

PR #641, opened 2020-07-24
Implement Safari CDP message handler

RedMickey:


src/debugger/cdp-proxy/CDPMessageHandlers/safariCDPMessageHandler.ts
```diff
@@ -1,35 +1,103 @@
...
+            dispatchDirection,
         };
     }
 
     public processApplicationCDPMessage(event: any): ProcessedCDPMessage {
-        let sendBack = false;
+        let dispatchDirection = DispatchDirection.FORWARD;
+        let communicationPreparationsDone = undefined;
+        if (!event.method || !event.method.match(/^Target/)) {
+            dispatchDirection = DispatchDirection.CANCEL;
```
SounD120, line 53: Why if method is not from `Target` domain we mark request as cancelled?

src/debugger/cdp-proxy/cordovaCDPProxy.ts
```diff
@@ -52,10 +54,13 @@ export class CordovaCDPProxy {
         this.logLevel = logLevel;
         this.logger = OutputChannelLogger.getChannel("Cordova Chrome Proxy", true, false, true);
         this.debuggerEndpointHelper = new DebuggerEndpointHelper();
+        this.browserInspectUri = args.webSocketDebuggerUrl || "";
         if (args.platform === "ios" && (args.target === "emulator" || args.target === "device")) {
```
SounD120, line 58: I suppose that it is better to use enum `TargetTyper` instead of string literals

src/debugger/cordovaDebugSession.ts
```diff
@@ -101,7 +114,13 @@ export class CordovaDebugSession extends LoggingDebugSession {
         this.cancellationTokenSource = new vscode.CancellationTokenSource();
         this.cdpProxyHostAddress = "127.0.0.1"; // localhost
         // this.terminateCommand = "terminate"; // the "terminate" command is sent from the client to the debug adapter in order to give the debuggee a chance for terminating itself
-        this.pwaChromeSessionName = "pwa-chrome"; // the name of Chrome debug session created by js-debug extension
+        if (session.configuration.platform === "ios"
```
SounD120, line 117: I noticed that for target and platform we have a lot checks with string literals. I suppose that it will be better to add enums for them to control platforms and targets easily.

src/debugger/cordovaDebugSession.ts
```diff
@@ -695,70 +727,67 @@ export class CordovaDebugSession extends LoggingDebugSession {
                 }
             };
 
-            const getSimulatorProxyPort = (packagePath): Q.IWhenable<{ packagePath: string; targetPort: number }> => {
+            const getSimulatorProxyPort = (iOSAppPackagePath): Q.IWhenable<{ iOSAppPackagePath: string; targetPort: number; iOSVersion: string }> => {
```
SounD120, line 730: Will it be useful if we place endpoint reply example here as a comment?

PR #645, opened 2020-07-31
Migrate unit tests from vscode to vscode-test package

RedMickey:


gulpfile.js
```diff
@@ -116,8 +116,8 @@ gulp.task('watch', gulp.series('build', function (cb) {
 }));
 
 gulp.task('run-test', function () {
-    return gulp.src('out/test/debugger/**/*.js', { read: false })
-        .pipe(mocha({ ui: 'bdd' }))
+    return gulp.src('out/test/debugger/**/*.test.js', { read: false })
```
SounD120, line 119: Maybe it is better to run only one entry file here? Or maybe we can do this for js-debug migration branch

PR #664, opened 2020-09-23
Fix simulator viewport handling

RedMickey:


src/debugger/cordovaDebugSession.ts
```diff
@@ -550,11 +547,6 @@ export class CordovaDebugSession extends LoggingDebugSession {
                     this.outputLogger(viewportResizeFailMessage, true);
                 }).done();
             });
-            this.simulateDebugHost.on("reset-viewport", () => {
-                this.resetSimulateViewport().catch(() => {
-                    this.outputLogger(viewportResizeFailMessage, true);
```
SounD120, line 555: I suppose that we should delete this message either.

SounD120 (approved):

Issue #594, opened 2019-12-04
Support custom cwd paths


Hadatko:
Hi i would like to see this feature in your great extension. I want to reduce amount of redundant directories in my workspace. 

## Actual Behavior
1. Not supported. CWD need be workspace root dir.

## Expected Behavior
1. CWD could be customed.

E.g. Google chrome debbuger support this:
"configurations": [{
"type": "chrome",
"request": "launch",
"name": "Launch Chrome against localhost",
"url": "http://localhost:8100",
**"webRoot": "${workspaceFolder}/gui/html"**,
"disableNetworkCache": true
}]

Thanks 

Reference: #591

haris44:
It can be really usefull, to use this vscode plugin with vue-cli cordova extensions for example :  https://github.com/m0dch3n/vue-cli-plugin-cordova

Path hierarchy with different folders (./ and ./src-cordova/) allow seperation between project node-modules and cordova plugin node-modules. I think it's a best practice with cordova project.  

Thanks

michaelsogos:
Hi @haris44 

I have same scenario, did you find a solution in the meantime?

haris44:
Hi @michaelsogos. yes, finally I used an other plugin : "Android Webview Debugging". It works really well ! (even with Chrome Webview)

JiglioNero:
The problem was solved in this [PR](https://github.com/microsoft/vscode-cordova/pull/667) and the fix will be released in the next version of the extension.

PR #667, opened 2020-09-25
Implement custom cwd support

JiglioNero:
#594


src/cordova.ts
```diff
@@ -77,12 +82,27 @@ function onChangeWorkspaceFolders(context: vscode.ExtensionContext, event: vscod
...
         });
     }
 }
 
-function onFolderAdded(context: vscode.ExtensionContext, folder: vscode.WorkspaceFolder): void {
+export function createAdditionalWorkspaceFolder(folderPath: string): vscode.WorkspaceFolder | null {
+    if (fs.existsSync(folderPath)) {
+        let projectRoot = folderPath;
+        if (path.isAbsolute(folderPath)) {
+            projectRoot = "/" + folderPath.charAt(0).toLowerCase() + folderPath.substring(1);
```
SounD120, line 94: Could you please provide an example of what this line does?

src/cordova.ts
```diff
@@ -33,26 +33,31 @@ let CORDOVA_TYPINGS_QUERYSTRING = "cordova";
...
 let TSCONFIG_FILENAME = "tsconfig.json";
 
+let EXTENSION_CONTEXT: vscode.ExtensionContext;
+let COUNT_WORKSPACE_FOLDERS = 9000;
+
 export function activate(context: vscode.ExtensionContext): void {
     // Asynchronously enable telemetry
     Telemetry.init("cordova-tools", customRequire(findFileInFolderHierarchy(__dirname, "package.json")).version, { isExtensionProcess: true, projectRoot: "" });
 
+    EXTENSION_CONTEXT = context;
```
SounD120, line 43: Are you sure that the context that VS Code sends during activation equals to the context which we can receive from other commands?

src/cordova.ts
```diff
@@ -77,12 +82,27 @@ function onChangeWorkspaceFolders(context: vscode.ExtensionContext, event: vscod
...
     if (event.added.length) {
         event.added.forEach((folder) => {
-            onFolderAdded(context, folder);
+            onFolderAdded(folder);
         });
     }
 }
 
-function onFolderAdded(context: vscode.ExtensionContext, folder: vscode.WorkspaceFolder): void {
+export function createAdditionalWorkspaceFolder(folderPath: string): vscode.WorkspaceFolder | null {
```
SounD120, line 90: I think that we should check that the added folder represents Cordova or Ionic project. Do we have that check?

src/cordova.ts
```diff
@@ -33,26 +33,31 @@ let CORDOVA_TYPINGS_QUERYSTRING = "cordova";
 let JSCONFIG_FILENAME = "jsconfig.json";
 let TSCONFIG_FILENAME = "tsconfig.json";
 
+let EXTENSION_CONTEXT: vscode.ExtensionContext;
+let COUNT_WORKSPACE_FOLDERS = 9000;
```
SounD120, line 37: Please explain why this number was chosen.

src/cordova.ts
```diff
@@ -198,7 +219,11 @@ function onFolderAdded(context: vscode.ExtensionContext, folder: vscode.Workspac
 }
 
 function onFolderRemoved(folder: vscode.WorkspaceFolder): void {
-    ProjectsStorage.delFolder(folder);
+    Object.keys(ProjectsStorage.projectsCache).forEach(path => {
+        if (CordovaProjectHelper.checkPathBelongsToHierarchy(folder.uri.fsPath.toLowerCase(), path)) {
+            delete ProjectsStorage.projectsCache[path];
```
SounD120, line 224: Maybe it is better to use delFolder function instead of removing directly?

SounD120 (approved):

PR #669, opened 2020-10-05
Add unit tests 

RedMickey:


src/cordova.ts
```diff
@@ -102,20 +102,15 @@ export function createAdditionalWorkspaceFolder(folderPath: string): vscode.Work
...
 export function onFolderAdded(folder: vscode.WorkspaceFolder): void {
     let workspaceRoot = folder.uri.fsPath;
-    let cordovaProjectRoot = CordovaProjectHelper.getCordovaProjectRoot(workspaceRoot);
 
-    if (!cordovaProjectRoot) {
-        return;
-    }
-
-    if (path.resolve(cordovaProjectRoot) !== path.resolve(workspaceRoot)) {
+    if (!CordovaProjectHelper.isCordovaProject(workspaceRoot)) {
```
SounD120, line 106: AFAIK this function can be called when the extension is already activated. This will mean that the customer will receive message that extensions hasn't been activated when it is already activated. Am I right?

src/utils/extensionHelper.ts
```diff
@@ -65,7 +65,6 @@ export function findFileInFolderHierarchy(dir: string, filename: string): string
     let atFsRoot: boolean = false;
 
     while (!fs.existsSync(path.join(projectRoot, filename))) {
-        // Navigate up one level until either config.xml is found
```
SounD120, line 68: What is the point removing this comment?

RedMickey: It seems that this was a redundant comment, since we don't search for the `config.xml` file here.

SounD120 (approved):

Issue #588, opened 2019-11-21
Are there any plans to add support for Edge Chromium?


EOussama:
## Actual Behavior
1. The debugging mode always launches **Chrome**.

## Expected Behavior
1. Being able to customize the preferred browser, from which **Edge Chromium** can be chosen.

SounD120:
Hi @EOussama and thanks for reaching us. When you are saying "customize the prefered browser" you mean add an ability to change default browser for simulate and serve debug scenarios?

EOussama:
> Hi @EOussama and thanks for reaching us. When you say "customize the prefered browser", do you mean adding an ability to change the default browser for simulate and serve debug scenarios?

Yes, exactly that.

SounD120:
Got it, thanks for your feedback, we'll consider that and manage in order of priorities.

kenicenoel:
Agreed. I only have Google Chrome installed for this reason as the new Edge doesn't work with Cordova tools. I want to be able to use the extension with Edge and not Google Chrome.

kenicenoel:
> Got it, thanks for your feedback, we'll consider that and manage in order of priorities.

I really hope this is prioritized as the new Edge is a Microsoft product and one would think their tooling should should be supported.

Namiro:
+1 I totaly agree with @kenicenoel

RedMickey:
Hi @EOussama, now Cordova Tools extension supports Edge Chromium browser for the `Serve to the browser (Ionic Serve)` debugging scenario. Simulate debugging scenarios (e.g. `"Simulate Android in browser`, `Run Browser`) haven't supported Edge Chromium yet, because they use the [`cordova-serve`](https://github.com/apache/cordova-serve) package, which doesn't support Edge Chromium, to run Cordova apps in the browser. We proposed [the PR](https://github.com/apache/cordova-serve/pull/40) to `cordova-serve` that adds support of Edge Chromium. So, when this PR is merged and a new version of the `cordova-serve` package is released, the extension should fully support Edge Chromium browser.

PR #684, opened 2020-10-20
Add MS Edge Chromium support

RedMickey:
https://github.com/microsoft/vscode-cordova/issues/588

### DIFF
--- a/src/extension/simulate.ts
+++ b/src/extension/simulate.ts
@@ -17,6 +17,8 @@ import * as nls from "vscode-nls";
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize = nls.loadMessageBundle();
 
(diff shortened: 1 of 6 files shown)


src/extension/simulate.ts
```diff
@@ -17,6 +17,8 @@ import * as nls from "vscode-nls";
 nls.config({ messageFormat: nls.MessageFormat.bundle, bundleFormat: nls.BundleFormat.standalone })();
 const localize = nls.loadMessageBundle();
 
+export const SIMULATE_TARGETS: string[] = ["default", "chrome", "chromium", "edge", "firefox", "ie", "opera", "safari"];
```
SounD120, line 20: AFAIK array is a ref type, so the array address in memory will be constant, but its content will not. It is better to use interface or enum for that since it will be constant for sure.

SounD120 (approved):

src/extension/simulate.ts
RedMickey: Fixed

EOussama:
So, did this close #588 ?

PR #686, opened 2020-10-21
Implement runtimeVersion support

### LINKED ISSUE #571: Add NVM Node.js version support
As for now VS Code [doesn't support debugging with NVM on macOS](https://github.com/Microsoft/vscode/issues/19951#issuecomment-422077768). But there is a [`runtimeVersion`](https://github.com/microsoft/vscode-node-debug/blob/fc36cb94a3a556803f5284a0e86a1e1564059076/package.nls.json#L48) debug configuration variable exists in [`vscode-node-debug`](https://github.com/microsoft/vscode-node-debug) to switch Node.js versions via nvm for debug sessions inside VS Code. We can investigate if we can implement it the [same way as in Node debugger](https://github.com/microsoft/vscode-node-debug/blob/fc36cb94a3a556803f5284a0e86a1e1564059076/src/node/extension/configurationProvider.ts#L204-L273).

SounD120:
Fixes #571


src/debugger/extension.ts
```diff
@@ -113,7 +113,7 @@ export function cordovaStartCommand(command: string, args: string[], env: any, c
         args.push("--no-update-notifier");
     }
 
-    return child_process.spawn(command, args, { cwd: cordovaRootPath, env });
+    return child_process.spawn(command, args, { cwd: cordovaRootPath, env: Object.assign(process.env, env) });
```
SounD120, line 116: Not sure if need to change anything here since the `env` that has been passed to this function might be already modified by our needs. This change might cause unpredictable consequences.

src/debugger/cordovaDebugSession.ts
```diff
@@ -1195,7 +1200,7 @@ To get the list of addresses run "ionic cordova run PLATFORM --livereload" (wher
         if (browserPath[0]) {
             this.browserProc = child_process.spawn(browserPath[0].path, chromeArgs, {
                 detached: true,
-                stdio: ["ignore"],
+                stdio: ["ignore"]
```
RedMickey, line 1203:
```suggestion
                stdio: ["ignore"],
```

src/debugger/extension.ts
```diff
@@ -15,7 +15,7 @@ const errorsToSuppress = ["Run an Ionic project on a connected device"];
 
 export function execCommand(command: string, args: string[], errorLogger: (message: string) => void): Q.Promise<string> {
     let deferred = Q.defer<string>();
-    let proc = child_process.spawn(command, args, { stdio: "pipe" });
+    let proc = child_process.spawn(command, args, { stdio: "pipe"});
```
RedMickey, line 18:
```suggestion
    let proc = child_process.spawn(command, args, { stdio: "pipe" });

```

src/utils/nodeVersionHelper.ts
```diff
@@ -0,0 +1,123 @@
...
+			}
+		}
+
+		if (!bin) {
+
+			// now try the Node Version Manager "nvm"
+			if (process.platform === "win32") {
+				const nvmHome = process.env["NVM_HOME"];
+				if (!nvmHome) {
+					throw new Error(localize("NVM_HOME.not.found.message", "Attribute 'runtimeVersion' requires Node.js version manager 'nvm-windows' or 'nvs'."));
```
SounD120, line 89: Hey @JiglioNero . Could you please change localize labels so they will suit our code style?
Example: `NVM_HOME.not.found.message` -> `nvmHomeNotFoundMessage`

RedMickey (approved):

PR #786, opened 2022-09-20
Refactor browser platform launching logic

EzioLi01:


EzioLi01:
Will look into this after adding edge support for MacOS and Linux platform.

